PR #3, opened 2022-09-08
Import channel implementations from Python SDK v0.9.0

shsms:


.darglint
```diff
@@ -0,0 +1,10 @@
+# -*- mode: ini -*-
```
leandro-lucarella-frequenz, line 1: Not a biggie, but can't you configure your editor locally to avoid having this kind of editor-dependent configuration here?

benchmarks/channel/benchmark_anycast.py
```diff
@@ -0,0 +1,131 @@
+"""Benchmark for Anycast channels.
+
+Copyright
+Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
+License
+MIT
```
leandro-lucarella-frequenz: :ok_hand:

pyproject.toml
```diff
@@ -0,0 +1,32 @@
+[build-system]
+requires = [
+    "setuptools >= 65.3.0, < 66",
+    "setuptools_scm[toml] >= 7.0.5, < 8",
+    "wheel"
+]
+build-backend = "setuptools.build_meta"
+
+[tool.pylint]
```
leandro-lucarella-frequenz: This *section* is not really needed

pyproject.toml
```diff
@@ -0,0 +1,32 @@
...
+    ignore-comments=['yes']
+    ignore-docstrings=['yes']
+    ignore-imports=['no']
+    min-similarity-lines=40
+
+    [tool.pylint.messages_control]
+    # disable wrong-import-order, ungrouped-imports because it conflicts with isort
+    disable = ["too-few-public-methods", "wrong-import-order", "ungrouped-imports"]
+    [tool.pylint.'DESIGN']
+    max-attributes=12
```
leandro-lucarella-frequenz: Nor the extra indentation. I never seen this styles in other projects or other TOML files, I'll try to keep it with a more widely used format.

setup.py
```diff
@@ -0,0 +1,8 @@
+"""
+Simple setuptools shim to allow editable installs from setup.cfg
+"""
```
leandro-lucarella-frequenz, line 8: Maybe add the copyright and license header here, even if it is a trivial file, just for consistency.

setup.cfg
```diff
@@ -0,0 +1,18 @@
...
+author = Frequenz Energy-as-a-Service GmbH
+
+[options]
+install_requires =
+    watchfiles >= 0.15.0
+
+package_dir =
+    = src
+packages = find_namespace:
+python_requires = >= 3.8, < 4
```
leandro-lucarella-frequenz, line 12: According to my experience, you need to also specify here:
```ini
include_package_data = True
```

Otherwise the `py.typed` file won't be included. You can try to build the distribution package and see if the file is included or not.

leandro-lucarella-frequenz (dismissed):
I didn't go through the code as it was imported, only looked at the supporting/tooling files.

I would also remove the extra `channels` sub-directory in both `benchmarks/` and `tests/`, as `channels` should be the only package provided by this repo.

Only the comment about `py.typed` is important to at least check if it is working. The rest are minor, so approving.

.darglint
shsms: This is really just to get syntax highlighting on github. Is there a different way to get syntax highlighting on gh for files with unknown extensions?

pyproject.toml
shsms: removed

pyproject.toml
shsms: this is updated as well.

setup.py
shsms: done

setup.cfg
shsms: no,  it works because we have:

``` ini
[options.package_data]
frequenz.channels = py.typed
```

shsms:
> I would also remove the extra channels sub-directory in both benchmarks/ and tests/, as channels should be the only package provided by this repo.

this is also done.

.darglint
leandro-lucarella-frequenz: Oh, I see. I didn't know about that. [git attributes](https://git-scm.com/docs/gitattributes) look like a very good match for this and it looks like they are supported (as well as both emacs local variables and VIM modelines):

https://github.com/github/linguist/blob/master/docs/overrides.md#using-gitattributes

I'm also fine if you want to leave it as it is, although I see an editor war coming... :boom: :laughing:

setup.cfg
leandro-lucarella-frequenz: Oh, damn, it uses python package syntax in here, not distribution package syntax! It didn't work for me in another project without `include_package_data = True` but I guess then probably that config makes `setuptools` include common known data and the actual `[options.package_data]` config was simply broken. :grimacing:

leandro-lucarella-frequenz (approved):

.darglint
shsms: Ok,  so we assign it to the first person that complains? (I mean you ;-))

setup.cfg
leandro-lucarella-frequenz: Just for completion:

`include_package_data = True` includes all the files that live on the package directory as long as:

1. These files are declared in `MANIFEST.in` (so they are part of the source distribution package)
2. OR, they are being tracked by a revision control system such as Git, Mercurial or SVN, and you have configured an appropriate plugin such as [setuptools-scm](https://pypi.org/project/setuptools-scm) or [setuptools-svn](https://pypi.org/project/setuptools-svn)

https://setuptools.pypa.io/en/latest/userguide/datafiles.html

So yeah, in our case one or the other is sufficient. I think I prefer to use `include_package_data = True` as we probably want to ship anything that we put in `src` and it is in git (otherwise it wouldn't be there). But I'm OK with both.

leandro-lucarella-frequenz: Just for completion:

`include_package_data = True` includes all the files that live on the package directory as long as:

1. These files are declared in `MANIFEST.in` (so they are part of the source distribution package)
2. OR, they are being tracked by a revision control system such as Git, Mercurial or SVN, and you have configured an appropriate plugin such as [setuptools-scm](https://pypi.org/project/setuptools-scm) or [setuptools-svn](https://pypi.org/project/setuptools-svn)

https://setuptools.pypa.io/en/latest/userguide/datafiles.html

So yeah, in our case one or the other is sufficient. I think I prefer to use `include_package_data = True` as we probably want to ship anything that we put in `src` and it is in git (otherwise it wouldn't be there). But I'm OK with both.

.darglint
leandro-lucarella-frequenz: That's **exactly** why I said I was fine with it :smile:

Seriously, I'm also not convinced to go with the attributes because then it is *linguist*-dependent. There seem to be no perfect solution for this, so any solution is fine.

leandro-lucarella-frequenz:
BTW, I'm trying to adopt a convention for "conversations" and using the :rocket: emoji to indicate for me we can *resolve* the conversation. I'm using it when I have a final comment because if I resolve it, then other people usually miss the comment because the conversation is collapsed. So if you see any conversations where my last message has a :rocket: emoji, feel free to mark it as resolved. If I don't have no further comments I will just resolve the conversations I started myself when whatever it was about is resolved.

shsms:
That's a nice idea 🚀

PR #19, opened 2022-10-05
ci: Test also Python 3.11 pre-releases

leandro-lucarella-frequenz:
We are very close to the release of Python 3.11 so it's a good idea to start testing aginst it too.


sahas-subramanian-frequenz (approved):
Super! We should also hook up some of the benchmarks to CI,  just to see how the perf improvements in python 3.11 will affect the channels. :crossed_fingers:

leandro-lucarella-frequenz:
Feel free to create an issue for that :)

PR #23, opened 2022-10-25
[RFC] Handle passing exceptions via channels

### LINKED ISSUE #21: Implement a general mechanism to send errors via channels
### What's needed?

A way to pass errors via channels.

### Proposed solution

One way to be able to pass exceptions through channels would be to make the channels aware of exceptions and if an exception instance is received via a channel, just raise an exception.

One problem with this approach is using `Select()` then becomes less clear, as we would need to wrap the `select.ready()` in a `try` / `except` block, which dissociates the error to the channel it appeared with.

It might be nicer to have a way to keep the handling of a particular channel (both good results and errors) in the same block. But then to get the data, we seem to always call `.inner` (not sure why), so that could be a good place to plug a check for the result sent to the channel and `raise` if it was an `Exception`.

### Use cases

`Bidirectional` channels often need to report errors if a *request* couldn't be fulfilled for example. Any error that comes up in a sender for which it might not be able to produce a proper message could also be sent to the receiver to let them know about the failure.

Define a message that is a `Union` of two types, the success message and an error message.

### Additional context

_No response_

leandro-lucarella-frequenz:
Exceptions are treated differently when passed via channels, and are re-raised on the receiving side.

The Select object is also adjusted to handle exceptions specially. Unless exceptions are handled explicitly (with a Future-like interface), they will be re-raised when trying to get a value from the channel.

This PR is very minimal on purpose, only to show the concept. It is incomplete and untested. Only the base classes, the AnycastChannel as an example and the Select object are adjusted.

I think it will make sense to make the closure of the channel an exception too, otherwise we have 2 error reporting mechanisms, so I will probably make it part of the same PR. But before doing that I will also make minimal changes to show the concept.

### DIFF
--- a/src/frequenz/channels/merge_named.py
+++ b/src/frequenz/channels/merge_named.py
@@ -11,11 +11,11 @@ import asyncio
 from collections import deque
 from typing import Any, Deque, Optional, Set, Tuple
 
-from frequenz.channels.base_classes import Receiver, T
+from frequenz.channels.base_classes import Message, Receiver, T
 
 
-class MergeNamed(Receiver[Tuple[str, T]]):
-    """Merge messages coming from multiple named channels into a single stream."""
+class MergeNamed(Receiver[Tuple[str, Message[T]]]):
+    """Merge messages coming from multiple named receivers into a single stream."""
 
     def __init__(self, **kwargs: Receiver[T]) -> None:
         """Create a `MergeNamed` instance.
@@ -28,15 +28,19 @@ class MergeNamed(Receiver[Tuple[str, T]]):
             asyncio.create_task(recv.receive(), name=name)
             for name, recv in self._receivers.items()
         }
-        self._results: Deque[Tuple[str, T]] = deque(maxlen=len(self._receivers))
+        self._results: Deque[Tuple[str, Message[T]]] = deque(maxlen=len(self._receivers))
 
     def __del__(self) -> None:
         """Cleanup any pending tasks."""
         for task in self._pending:
             task.cancel()
 
-    async def receive(self) -> Optional[Tuple[str, T]]:
-        """Wait until there's a message in any of the channels.
+    async def receive(self) -> Optional[Tuple[str, Message[T]]]:
+        """Wait until there's a message in any of the merged receivers.
+
+        If any exception was received via the merged receivers it will NOT be
+        raised, it will be returned instead, so the caller should check in any
+        of the received message is an exception.
 
         Returns:
             The next message that was received, or None, if all channels have
@@ -55,7 +59,32 @@ class MergeNamed(Receiver[Tuple[str, T]]):
             )
             for item in done:
                 name = item.get_name()
-                result = item.result()
+                # TODO: Check if we can first put back all done tasks into
+                # self._pending so they are picked up on the next receive()
+                # call.
+                #
+                # Otherwise we could wrap the messages like it is done in
+                # `Select`, so the user would call wrapped.result and that
+                # would raise, so raising can be delayed to the point where the
+                # value needs to be unwrapped.
+                #
+                # For now we are simply not raising any received exceptions.
+                #
+                # Raising a received exception here would mean messages can be
+                # lost for the following done items. Because of this, we
+                # collect the exceptions instead and return them.
+                if exception := item.exception():
+                    # Only Exceptions should come in the receiver, so we
+                    # re-raise any other exception (BaseException) as it should
+                    # be a programming error.
+                    #
+                    # Also we need this for type-checking, as task.exception is
+                    # Optional[BaseException] and we want to store this in
+                    # a Message[T], which is a Union[T, Exception].
+                    assert isinstance(exception, Exception)
+                    result = exception
+                else:
+                    result = item.result()
                 # if channel is closed, don't add a task for it again.
                 if result is None:
                     continue
--- a/src/frequenz/channels/select.py
+++ b/src/frequenz/channels/select.py
@@ -20,15 +20,49 @@ logger = logging.Logger(__name__)
 T = TypeVar("T")
 
 
-@dataclass
+@dataclass(frozen=True)
 class _Selected:
     """A wrapper class for holding values in `Select`.
 
     Using this wrapper class allows `Select` to inform user code when a
-    receiver gets closed.
+    receiver gets closed or raises an exception.
     """
 
-    inner: Optional[Any]
+    _result: Any
+
+    @property
+    def result(self) -> Any:
+        """The result from a selected receiver or None if the channel was closed.
+
+        Returns: the value of the receiver or None if the channel was closed.
+
+        Raises:
+            Exception: if the value is an exception.
+        """
+        if isinstance(self._result, Exception):
+            raise self._result
+        return self._result
+
+    @property
+    def exception(self) -> Optional[Exception]:
+        """The exception from a selected receiver.
+
+        Returns: the exception that was received or None if there was no exception.
+
+        Raises:
+            Exception: if the value is an exception.
+        """
+        if isinstance(self._result, Exception):
+            return self._result
+        return None
+
+    @property
+    def closed(self) -> bool:
+        """Whether the underlaying channel from the selected receiver was closed.
+
+        Returns: True if it was closed, False if it wasn't.
+        """
+        return self._result is None
 
 
 class Select:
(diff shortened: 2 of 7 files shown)


src/frequenz/channels/select.py
```diff
@@ -20,15 +20,49 @@
...
+        """The result from a selected receiver or None if the channel was closed.
+
+        Returns: the value of the receiver or None if the channel was closed.
+
+        Raises:
+            Exception: if the value is an exception.
+        """
+        if isinstance(self._result, Exception):
+            raise self._result
+        return self._result
```
sahas-subramanian-frequenz, line 44: I guess this should be `value` instead of `result`?  Result would suggest some operation was performed on some other data to produce the result.

Also,  why not just have `_result` and `_exception` as instance variables?  Else,  when users call `exception` first, we'd do an `if isinstance`, and once again when they call `result`.  Not a big cost,  but using variables would also mean simpler code as well.

sahas-subramanian-frequenz (commented):
I guess when we implement channels in a language with exceptions,  the channels need to support exceptions too.  It looks fairly non-intrusive as well.  And the idea of using an exception to signal channel closure makes the `None` value available as a regular value,  so channels can actually send `None` values.

src/frequenz/channels/select.py
leandro-lucarella-frequenz: > I guess this should be value instead of result? Result would suggest some operation was performed on some other data to produce the result.

Yeah, I wasn't sure, I like that this follows the `Future` interface, but it is not really a future (so the interface is much more limited than a `Future`). We could name it `value`. I was also thinking about `message`. It seems we are not using very consistent naming, sometimes it says *value*, sometimes *message*.

Maybe we can say a *message* can carry a *value* or an *exception*?

sahas-subramanian-frequenz: > Maybe we can say a message can carry a value or an exception?

yes,  that makes sense.

leandro-lucarella-frequenz: > Also, why not just have _result and _exception as instance variables? Else, when users call exception first, we'd do an if isinstance, and once again when they call result. Not a big cost, but using variables would also mean simpler code as well.

I could do that too, I have no strong reason, I was just mimicking what comes in the channel. I guess the most typical use case will be using `result` and trying to catch the exception and `result` will need to check if there was no exception anyway.

sahas-subramanian-frequenz: oh I see,  `result` raises.  ok,  then its probably better like this.

leandro-lucarella-frequenz:
I pushed a new commit with changes to raise an exception when the channel is closed (so allowing passing `None` via channels too). While doing so I found myself converting exception back and forth between `StopAsyncIteration` and `ChannelClosedError` and it got me thinking why do we even have `receive()` in the `Receiver` and why not just `Receiver = AsyncIterator`?

And that makes me think if channels shouldn't be `AsyncIterable`s, so that would be the way to get a receiver? But this one might be going too far, it can be a bit weird to use `receiver = aiter(chan)` instead of `receiver = chan.get_receiver()` (and also `aiter` is only available in Python 3.10+).

I was also a bit surprised that we have no interface for channels defined, because we seem to have some common interface (at the minimum `get_receiver()` and `get_sender()`). I'm still unsure if I feel good or bad about it. :D

When writing this I also had a few other questions about how to implement it but now I forgot :-/

Well, one of the questions I made to myself was if we should allow passing `BaseException`s or not via channels. I guess not, it would be weird, but I bring it up in case you have any opinions about it.

Related to using message for (value or exception), I think I tried to define a type alias `Message = Union[T, Exception]` to avoid typing that everywhere but it didn't work, I think type aliases can't be done parametrized types.

So, in general, besides the small comments, for you is a :+1: for making a proper PR for the first commit (sending exceptions via channels), is that correct? Because it is I can work on it.

Let me know what do you think about the second part (except for closed channels/allow passing `None` via channels).

leandro-lucarella-frequenz:
I remembered a couple of more questions that came to mind when deciding on the design.

At some point I thought about adding a `send_exception()` function to sender, to make it more clear that we are sending an exception and not a regular value. But if we do that, for me it would be weird that an `Exception` object sent via the regular `send()` would be raised at the receiving end. If we really want to have sort of 2 separate channels in one, one for values and one for exceptions, then we'll need to wrap every message passed through a channel and have 2 separate attributes (or a bool) to distinguish is something is intended to be raised at the receiving end or not, which seemed a bit overkill. Specially considering that it would be very weird to want to send an `Exception` that doesn't have to be raised at the receiving end, and in the weird case one might really need that, it can be wrapped manually by the user (so sending, for example, `(exception,)` or a `dataclass` with one field `exception` instead of `Exception`; the receiving end will see the message is a tuple and not raise).

As you said, I find the current approach much less intrusive and still with a reasonable interface and flexible enough that if you need to do crazy things, they might look ugly, but they are still doable.

sahas-subramanian-frequenz:
> While doing so I found myself converting exception back and forth between StopAsyncIteration and ChannelClosedError and it got me thinking why do we even have receive() in the Receiver and why not just Receiver = AsyncIterator?

The send and recieve methods are inspired by the rust implementation of channels,  and I think they complement each other and it makes sense to have them to maintain the illusion of a channel.  But maybe we can switch around the order of the messages - with `__anext__` doing the actual work,  and then `receive` calling `__anext__`?

> I was also a bit surprised that we have no interface for channels defined, because we seem to have some common interface (at the minimum get_receiver() and get_sender()). I'm still unsure if I feel good or bad about it. :D

I thought that this would encourage the passing around of channels,  instead of just senders and receivers.  And things like `Bidirectional` won't fit into such an abstraction anyway.

> Well, one of the questions I made to myself was if we should allow passing BaseExceptions or not via channels. I guess not, it would be weird, but I bring it up in case you have any opinions about it.

I have no opinions,  but why is it weird to pass `BaseException`s?

> Let me know what do you think about the second part (except for closed channels/allow passing None via channels).

yes,  this is a very nice idea,  taking away the special meaning of `None` values.

> Specially considering that it would be very weird to want to send an Exception that doesn't have to be raised at the receiving end, and in the weird case one might really need that, it can be wrapped manually by the user.

a channel for sending Exception objects? :D then they can just make a try block around the calls to receive.  It works in go because errors are just values,  with no special meaning.  We don't have to provide direct support for it in Python.

leandro-lucarella-frequenz:
> The send and recieve methods are inspired by the rust implementation of channels, and I think they complement each other and it makes sense to have them to maintain the illusion of a channel. But maybe we can switch around the order of the messages - with `__anext__` doing the actual work, and then `receive` calling `__anext__`?

Yes, we could. It will probably at least save one or two conversions between `StopAsyncIterator` and `ChannelClosedError`.

> I thought that this would encourage the passing around of channels, instead of just senders and receivers. And things like `Bidirectional` won't fit into such an abstraction anyway.

:+1: 

> I have no opinions, but why is it weird to pass `BaseException`s?

Because they are exceptions that are supposed to terminate the program (and are not caught by `except Exception`, which is the recommended way to catch exceptions). It would be more like a `panic` in Go I guess. Only `SystemExit`, `KeyboardInterrupt` and `GeneratorExit` inherit from `BaseException`.

> a channel for sending Exception objects? :D then they can just make a try block around the calls to receive. It works in go because errors are just values, with no special meaning. We don't have to provide direct support for it in Python.

Good, so I guess we are on the same page with this one, I just wanted to make sure we are all on the same page.

> yes, this is a very nice idea, taking away the special meaning of `None` values.

Good, then I take from this that I should proceed to finish the implementation of both passing exceptions via channels and using an exception to signal a channel was closed as drafted in this PR.

sahas-subramanian-frequenz:
> Good, then I take from this that I should proceed to finish the implementation of both passing exceptions via channels and using an exception to signal a channel was closed as drafted in this PR.

yup,  and probably also the switching around of `__anext__` and `receive`.

src/frequenz/channels/merge_named.py
```diff
@@ -55,7 +59,32 @@ async def receive(self) -> Optional[Tuple[str, T]]:
...
+                    # re-raise any other exception (BaseException) as it should
+                    # be a programming error.
+                    #
+                    # Also we need this for type-checking, as task.exception is
+                    # Optional[BaseException] and we want to store this in
+                    # a Message[T], which is a Union[T, Exception].
+                    assert isinstance(exception, Exception)
+                    result = exception
+                else:
+                    result = item.result()
```
leandro-lucarella-frequenz, line 87: This is still on the works but I just bumped into this case, which is not as unintrusive as I would have liked. I will probably have a look at the solutions I'm mentioning there, but though it might be a good idea to bring it up as soon as possible in case you have other ideas.

leandro-lucarella-frequenz:
After some internal discussion, we agreed on not implementing raising exceptions when passed via channels. I will add more details to the issue.

PR #25, opened 2022-10-31
Use MkDocs to build the documentation

leandro-lucarella-frequenz:
The documentation is generated using MkDocs, using the Material theme plus mkdocstrings (and a few minor supporting extensions) to build the API documentation extracted from code comments.

For now the home page is a place holder, and the supporting script `docs/mkdocstrings-autoapi.py` should eventually be moved to a common repository.

### TODO

- [x] Build and publish the docs via GitHub Pages in the CI.
- [x] Generate site for each version using `mike`.
- [x] Do some basic documentation fixing, like using admonitions for `Copyright` and `License`.

#### Probably for another PR

- [x] Think about what is the public interface, and hide modules that are internal (probably we only want to publish what's exported via `frequenz.channels`). #30 
- [x] Add class attributes documentation. #31
- [ ] Move documentation tips (section above) to some more permanent place so it doesn't get lost.
- [x] Generate docs for the v0.10.0 release
- [x] Configure the default version in `mike` on the final website.
- [x] Look for failing deployment (permissions)

### Notes on how documentation should be written:

- Properties:
  - `@properties` are rendered as function (with a special mark for being a property), but this is planned to be improved in the future to be shown as attributes (see https://github.com/frequenz-floss/frequenz-channels-python/pull/25#issuecomment-1309059829).
  - Removing documentation from the function and using `Attributes:` instead somewhat works, but you only get a table in the `Class documentation`, attributes are not shown in the right ToC, which makes them hard to discover IMHO.
  - The best approach to `@properties`, that works well with `pydocstyle` and `darglint` seems to be the following:

    ```py
    class AClass:
        """A class."""

        @property
        def a_prop(self) -> int:
            """Get the a prop.

            Returns:
                The a prop of the class.
            """
            return 1

        @a_prop.setter
        def a_prop(self, a: int) -> None:
            pass
    ```

    image

    The only downside is properties are still listed as functions, not attributes (but this will change, so all good with this), and there is redundancy in the documentation (basically we need to document it twice, one for the description and one for the `Returns:`. `mkdocstrings` doesn't show the type if `Returns:` is not specified, and if the description is not included, `pydocstyle` complains.

- Class / instance attributes:
  - Attributes should be documented in general using *docstrings*, not the `Attributes:` section, as the `Attribute:` section doesn't produce ToC entries and doesn't extract a lot of info.

    ```py
    class A:
        """A class.

        Attributes:
            some_attr_in_attributes: Attribute in `Attributes:`, shows nice in the
                docs (the type is shown, no value is shown) but no ToC.
        """

        class_attr: str
        "This is marked as class attribute, not good."

        def __init__(self, t: tuple[int, int]):
            self.some_attr_in_attributes: tuple[int, int] = t

            self.some_docstr_attr_with_init: tuple[int, int] = t
            "This is marked as a instance attribute, but the value is shown out of context, not good."

            self.some_docstr_attr_without_init: tuple[int, str]
            "If we separate declaration from initialization, the declaration is not show, not good."

            self.some_docstr_attr_without_init = (2, "a")

            self.some_docstr_attr_without_init_doc_in_assign: tuple[int, str]
            self.some_good_attr = (2, "a")
            "But if we document the assignment, it is shown, but no type and out of context value, not good."
    ```
    
    image

  So it looks like there is no perfect solution, but using the style of `some_docstr_attr_with_init` seems to be the best approach.

- Module attributes / type aliases

  I tried this:

  ```py
  """Module docs.

  Attributes:
      SomeType: This is a type with lala.
  """

  SomeType = str
  ```

  And it produces:

  image

  While:
  ```py
  """Module docs."""

  SomeType = str
  """This is a type with lala."""
  ```

  And it produces:

  image

  So the former is definitely the way to go.

- To make cross-reference links we need to use this format `[some_symbol[fully.qualified.symbol.name]`.
- We should consider to make only the `frequenz.channels` documentation public, as the rest is not intended for end users and just makes the documentation harder to navigate (and could lead to users using inner modules, like `frequenz.channels.select.Select` instead of simply `frequenz.channels.Select`. We could also build 2 sets of API references: one for users and one for developers/collaborators.
- We need to use `from __future__ import annotations` and use the symbols instead of strings, otherwise the documentation doesn't get the cross references for these types.
- Examples should be put indented and at the end of the documentation for better rendering:

  ```py
  """Doc doc doc

  Doc doc doc.

  Example:
      ```python
      print("test")
      ```
  """
  ```
- There is a difference between `Example:` and `Examples:` (plural). The former creates an *admonition* and the later is used to write Python REPL examples directly, for example:

  ```py
  """Doc doc doc

  Doc doc doc.

  Examples:
      >>> 1 + 1
      2
      ```
  """
  ```

Two other minor commits with fixes are included:

- ci/labeler: Sort file patterns
- ci/labeler: Add missing noxfile.py


sahas-subramanian-frequenz:
I'm unable to build it locally with Python 3.10, with this error:

```
INFO     -  Cleaning site directory
INFO     -  Building documentation to directory: /home/sahas/code-oss/frequenz-channels-python/site
reference/frequenz/channels/index.md src/frequenz/channels/__init__.py
reference/frequenz/channels/anycast.md src/frequenz/channels/anycast.py
reference/frequenz/channels/base_classes.md src/frequenz/channels/base_classes.py
reference/frequenz/channels/bidirectional.md src/frequenz/channels/bidirectional.py
reference/frequenz/channels/broadcast.md src/frequenz/channels/broadcast.py
reference/frequenz/channels/merge.md src/frequenz/channels/merge.py
reference/frequenz/channels/merge_named.md src/frequenz/channels/merge_named.py
reference/frequenz/channels/select.md src/frequenz/channels/select.py
reference/frequenz/channels/utils/index.md src/frequenz/channels/utils/__init__.py
reference/frequenz/channels/utils/file_watcher.md src/frequenz/channels/utils/file_watcher.py
reference/frequenz/channels/utils/timer.md src/frequenz/channels/utils/timer.py
ERROR    -  mkdocstrings: frequenz.channels could not be found
ERROR    -  Error reading page 'reference/frequenz/channels/index.md':
ERROR    -  Could not collect 'frequenz.channels'

Aborted with a BuildError!
```

Do you know why that is?  Also,  what do you think about adding a test in CI that ensures the docs are building?

leandro-lucarella-frequenz:
> ERROR    -  mkdocstrings: frequenz.channels could not be found

Did you make a local install via `pip install -e .[docs]`?

> Do you know why that is? Also, what do you think about adding a test in CI that ensures the docs are building?

Yeah, it is in the plan.

leandro-lucarella-frequenz:
Updated with some fixes to the docs and added some notes I took on things to watch out for when writing documentation in the future. FYI @ela-kotulska-frequenz too.

docs/mkdocstrings-autoapi.py
```diff
@@ -1,3 +1,6 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+
 '''Generate the code reference pages.
```
sahas-subramanian-frequenz: Should we run this script through `black` as well?  The single quotes look a bit unplesant.

benchmarks/benchmark_broadcast.py
```diff
@@ -1,10 +1,7 @@
...
-
-Copyright
-Copyright © 2022 Frequenz Energy-as-a-Service GmbH
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
-License
-MIT
+"""Benchmark for Broadcast channels.
 """
```
sahas-subramanian-frequenz: Doesn't `black` automatically reformat these into single line comments? :open_mouth:

src/frequenz/channels/utils/timer.py
```diff
@@ -66,16 +66,18 @@ def reset(self) -> None:
...
         """
         self._stopped = True
 
     async def receive(self) -> Optional[datetime]:
         """Return the current time once the next tick is due.
 
         Returns:
-            The time of the next tick.
+            The time of the next tick or `None` if it was
+                [stop()][frequenz.channels.Timer.stop]ed.
```
sahas-subramanian-frequenz: The `stop()`ed looks a bit odd to me.  Do you like one of these better?

```
or `None` if `stop()` has been called on the timer
```

or 

```
or `None` if the timer has been [stopped][frequenz.channels.Timer.stop]
```

src/frequenz/channels/merge.py
```diff
@@ -14,16 +14,16 @@
 class Merge(Receiver[T]):
     """Merge messages coming from multiple channels into a single stream.
 
-    For example, if there are two channel receivers with the same type, they
-    can be awaited together, and their results merged into a single stream, by
-    using `Merge` like this:
+    If there are two channel receivers with the same type, they can be awaited
+    together, and their results merged into a single stream.
```
sahas-subramanian-frequenz: This is a specialization of the first line in the docstring,  and is suitable only for the example,  not as a general description.  I think it makes sense to move this paragraph under the `Example` section as well.

sahas-subramanian-frequenz:
> Did you make a local install via `pip install -e .[docs]`?

I did.  I followed the steps in the instructions,  and I still get the error.

benchmarks/benchmark_broadcast.py
leandro-lucarella-frequenz: Nope. Fixed.

docs/mkdocstrings-autoapi.py
leandro-lucarella-frequenz: We could. My plan is to move it away on its own repo eventually, as we'll need to reuse it, but it is true that until that happens it is good to have it checked.

src/frequenz/channels/utils/timer.py
leandro-lucarella-frequenz: I went for the first option :+1:

leandro-lucarella-frequenz:
> I did. I followed the steps in the instructions, and I still get the error.

This is super weird. I built a venv from scratch and did a `pip install -e .[docs]` and it works perfectly. From the error you get it seems like `mkdocs` can't find the `frequenz.channels` package, which should only point to `.` not being installed :thinking:

src/frequenz/channels/merge.py
leandro-lucarella-frequenz: I left it as before but inside the example.

docs/mkdocstrings-autoapi.py
leandro-lucarella-frequenz: I did, and now checking the `noxfile.py` too with improved `mypy` checks :rocket:

leandro-lucarella-frequenz:
All comments addressed and I added versioning support for the website. You can see an example here: https://leandro-lucarella-frequenz.github.io/frequenz-channels-python/

image

leandro-lucarella-frequenz:
Removing the draft status, as this should be ready to merge after the review.

pawamoy:
Chiming in again :sweat_smile: 

In docstrings, `Example:` sections (singular) will be rendered as admonitions, while `Examples:` sections (plural) will be parsed by mkdocstrings to support regular Python REPL examples, for example:

```
Examples:
    >>> print(1 + 1)
    2
```

leandro-lucarella-frequenz:
> Chiming in again

You are always welcome!

> In docstrings, Example: sections (singular) will be rendered as admonitions, while Examples: sections (plural) will be parsed by mkdocstrings to support regular Python REPL examples, for example:

OK, this one is unexpected. Did I missed something in the mkdocstrings documentation again or it is an undocumented feature?

While I have your attention, I wonder if there are any plans or reasons not to include `Attributes:` of a class in the ToC sidebar? Also what would be your recommended way to document `@properties` (as functions or as `Attributes:`)?

Thanks!

leandro-lucarella-frequenz:
I added how to document class / instance attributes and module attributes to the "Notes on how documentation should be written:" section in the issue.

I still have to test how property setters are handled.

sahas-subramanian-frequenz:
> I built a venv from scratch and did a pip install -e .[docs] and it works perfectly. 

works for me in a venv as well.  looks like my python env is broken.  I will look into it.

The PR looks good to me,  if you want to take this in and deal with properties next.

leandro-lucarella-frequenz:
Yeah, I would merge first and then do further fixing, I think that's lower prio. It is important to know how should we do the documentation though, so we document new code in a correct way, but fixing old stuff is lower prio IMHO. Will do a rebase and then it is ready to merge. We'll need to find a way to put the documentation guidelines/tips I collected above.

sahas-subramanian-frequenz (approved):
:tada:

leandro-lucarella-frequenz:
OK, the deployment is failing because of permissions problems (I didn't have this in my fork because it is less restrictive). I will look into it tomorrow.

pawamoy:
@leandro-lucarella-frequenz 

> While I have your attention, I wonder if there are any plans or reasons not to include Attributes: of a class in the ToC sidebar? 

Attributes sections, just like Parameters, Returns, Yields, etc., are rendered as tables, *without any particular heading*. The absence of headings means there will be no entry in the Table of Contents.

If you do want to get a heading for each documented attribute, you have to document the attribute itself rather than describing it in its parent class or module:

```python
SOME_ATTR: int = 0
"""Description of this attribute."""
```

It boils down to preference: are the attributes important enough that they should have a dedicated heading (and therefore permalink, which is quite useful I admit), or can we just have it described succinctly in the parent object :confused: 

> Also what would be your recommended way to document @properties (as functions or as Attributes:)?

Currently, properties are fetched as functions (because that's how they are declared), but I do not like this and am planning on returning them as attributes instead, since that's how they *accessed* (which is more important than how they are *declared*, since we're documenting API usage). It's in the backlog :smile: In the meantime, properties are rendered as functions, with trailing parentheses :confused: IIRC, properties setter and deleters are simply not taken into account, but we have plans for them as well, something like the ability to merge their docstrings together.

PS: we have a valuable conversation here, I might re-use your questions and my answers in the docs since I think it would help other users :slightly_smiling_face: Thanks!

leandro-lucarella-frequenz:
> It boils down to preference: are the attributes important enough that they should have a dedicated heading (and therefore permalink, which is quite useful I admit), or can we just have it described succinctly in the parent object confused

Yeah, for me the table is not an option because I think the visibility gets drastically reduced. At least my expectation from an API documentation is to find attributes in the TOC. It's what happens in most other languages documentation systems.

> Currently, properties are fetched as functions (because that's how they are declared), but I do not like this and am planning on returning them as attributes instead, since that's how they accessed (which is more important than how they are declared, since we're documenting API usage).

Yeah, 100%, for the user reading the docs it should be transparent if the implementation changes from a pure attribute to a property and vice-versa.

> It's in the backlog smile In the meantime, properties are rendered as functions, with trailing parentheses confused IIRC, properties setter and deleters are simply not taken into account, but we have plans for them as well, something like the ability to merge their docstrings together.

:tada: :tada: :tada: 

Good to know what's the future direction so we can document things properly and when things get improved we'll get the improvements without extra effort.

Thanks a lot for sharing, it is really useful stuff!

> PS: we have a valuable conversation here, I might re-use your questions and my answers in the docs since I think it would help other users slightly_smiling_face Thanks!

Great, glad to be of help too!

PR #28, opened 2022-11-03
Remove `version:xxx` and add `part:select` and `part:receivers` labels

leandro-lucarella-frequenz:
`version:xxx` was never really used and it is really hard to maintain as labels (and these affected files) need to be updated each time a version is released.

We also add new the new `part:xxx` labels.


leandro-lucarella-frequenz:
(enabled auto-merge, so approve only if you are OK with merging too)

sahas-subramanian-frequenz (approved):

PR #33, opened 2022-11-04
ci: Add contents permission to publish-docs

leandro-lucarella-frequenz:
Without this permissions the job can't publish GitHub Pages.


leandro-lucarella-frequenz:
I also had to add a specific *branch protection* for the branch `gh-pages` so it can be pushed without going through a PR. Still force-pushes and deletion are not allowed. At some point it might be good to create a special CI user and only allow pushing to this branch without PRs to this user, but not something to do right now.

sahas-subramanian-frequenz (approved):

PR #34, opened 2022-11-04
Add PyPI and Docs badges to README

shsms:
There's an example of what this would look like here:  https://github.com/shsms/frequenz-channels-python/blob/readme-badges/README.md

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,5 +1,9 @@
 # Frequenz channels
 
+[<img alt="GitHub Workflow Status" src="https://img.shields.io/github/workflow/status/frequenz-floss/frequenz-channels-python/frequenz-channels-python">](https://github.com/frequenz-floss/frequenz-channels-python/actions/workflows/ci.yaml)
+[<img alt="PyPI" src="https://img.shields.io/pypi/v/frequenz-channels">](https://pypi.org/project/frequenz-channels/)
+[<img alt="PyPI" src="https://img.shields.io/badge/docs-latest-informational">](https://frequenz-floss.github.io/frequenz-channels-python/)
+
 This repository contains channel implementations for python.
 
 ## Contributing


README.md
```diff
@@ -1,5 +1,8 @@
 # Frequenz channels
 
+[<img alt="GitHub Workflow Status" src="https://img.shields.io/github/workflow/status/frequenz-floss/frequenz-channels-python/frequenz-channels-python?label=docs">](https://frequenz-floss.github.io/frequenz-channels-python/)
```
leandro-lucarella-frequenz: I the webpage to the project, not sure if it is necessary to add it here too, but if we do it was a bit confusing to me that you get a GitHub CI badge (that covers all the CI, not just the docs) and then you get pointed to the website.

Also even if the docs build are currently failing you can still go to the docs and you won't see the docs for the default branch by default, you'll see the docs for the latest release, so is not like if they are failing you can't see the docs.

Maybe it is better to have a regular GitHub actions badge for the whole CI and just a link to the docs?

I can also go with this though, they are just random thoughts.

shsms: If we had a separate workflow for CI and release,  we could use the release workflow for this.  We can of course a static badge as well.  Let me make one just to see.

sahas-subramanian-frequenz: How does this look: https://github.com/shsms/frequenz-channels-python/blob/readme-badges/README.md

leandro-lucarella-frequenz: Yeah, I tried doing that, but I had to duplicate some steps, which wasn't ideal. There is probably a way to do it if we can use one workflow result in another. At the time I tried it briefly but it was complicated so I preferred to keep it simple for now. But I agree it would be ideal to have more independent workflows than one giant with everything.

leandro-lucarella-frequenz (dismissed):
Yeah, this definitely looks better, thanks!

leandro-lucarella-frequenz:
(please remember to assign the milestone :innocent: )

leandro-lucarella-frequenz (approved):

PR #35, opened 2022-11-04
Backport documentation generation to v0.10.x

leandro-lucarella-frequenz:
See #25 for details.

- ci/labeler: Add missing noxfile.py
- Use MkDocs to build the documentation
- Move copyright and license as code comments
- docs: Add missing file format to examples
- docs: Add some useful cross-references.
- docs: Indent examples
- Use forward type annotations
- docs: Indent "Returns:" properly
- docs: Improve formatting of some symbols
- docs: Make some minor improvements
- docs: Add site version via mike
- ci: Remove unnecessary checkout of submodules
- ci: Build and publish documentation
- Run normal linting on noxfile.py
- Run mypy on noxfile.py and be more restrictive
- ci: Add contents permission to publish-docs


sahas-subramanian-frequenz:
Does this push to the `gh-pages` branch as well?  Would it write to a different directory, so that it doesn't overwrite the docs for the other versions?

leandro-lucarella-frequenz:
> Does this push to the `gh-pages` branch as well? Would it write to a different directory, so that it doesn't overwrite the docs for the other versions?

Yes, this is what `mike` is for. It basically checks-out the current `gh-pages` branch, generates the docs in the directory specified by `<version>` in `mike deploy <version>`, overwriting that version but without touching other directories, so documentation for other versions will be left untouched, then commits. If `--push` is used, it will also push, but we could also push manually using git.

Then it uses *aliases* for versions. I'm using `latest` to point to the last stable release, which is also the default (what you get if you go to the `/` without pointing to a specific version). I'm also using `v0` to point to the latest `v0.x` and generating only vMAJOR.MINOR versions (so no specific docs for point releases).

`mike set-default <version>` just creates a `index.html` with a redirection. You can see this in the [`gh-pages` branches history](https://github.com/frequenz-floss/frequenz-channels-python/commits/gh-pages); here is the commit setting the default: https://github.com/frequenz-floss/frequenz-channels-python/commit/27dab7bcffc91ddb0b25b9290c65b577530069ec.

leandro-lucarella-frequenz:
FYI one check is missing because we didn't support Python 3.11-dev in v0.10.x, I don't think it is worth changing. Maybe we can add a dummy ci job that depends on all the tests we want the PRs to pass and only have that job as a requirement for PRs, so this doesn't happen while we add/remove python versions...

sahas-subramanian-frequenz:
>  because we didn't support Python 3.11-dev

we also need to switch to the stable version of 3.11, now that it has been released.

leandro-lucarella-frequenz:
> we also need to switch to the stable version of 3.11, now that it has been released.

Yeah, I was planning to do so after all this.

sahas-subramanian-frequenz (approved):

PR #38, opened 2022-11-04
ci: Update aliases when deploying with mike

leandro-lucarella-frequenz:
This is needed because we expect aliases we are deploying to already exist, and mike by default will only create new aliases, but not "move" them.

For example, if the current stable is v0.10.0 (`mike` version will be `v0.10` and `latest` and `v0` will point to it) and we just released v0.11.0, then the `latest` and `v0` aliases will need to be changed to point to `v0.11`.

For more details see:
https://github.com/jimporter/mike#building-your-docs

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -174,7 +174,7 @@ jobs:
           VERSION: ${{ steps.mike-metadata.outputs.version }}
           ALIASES: ${{ steps.mike-metadata.outputs.aliases }}
         run: |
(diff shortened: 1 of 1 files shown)


leandro-lucarella-frequenz:
I discover this while deploying the v0.10.1 version, as the `latest` alias needed to be changed from `next` to `v0.10`.

leandro-lucarella-frequenz:
(setting auto-merge, so only approve if you are OK with merging right away!)

sahas-subramanian-frequenz (approved):

Issue #27, opened 2022-11-03
Make `Select.ready` report one event at the time


leandro-lucarella-frequenz:
### What's needed?

When using `Select.ready()` it is a bit unexpected that more than one *event* could be received, for example:

```py
            while await select.ready():
                if _ := select.resampling_timer:
                    awaitables = [
                        self._output_senders[channel_name].send(sample)
                        for channel_name, sample in self._resampler.resample()
                    ]
                    await asyncio.gather(*awaitables)
                if msg := select.component_data_receiver:
                    if msg.inner is None:
                        # When this happens, then DataSourcingActor has closed the channel
                        # for sending data for a specific `ComponentMetricRequest`,
                        # which may need to be handled properly here, e.g. unsubscribe
                        continue
                    channel_name, sample = msg.inner
                    if self.is_sample_valid(sample=sample):
                        self._resampler.add_sample(
                            time_series_id=channel_name,
                            sample=sample,
                        )
                if msg := select.subscription_receiver:
                    if msg.inner is None:
                        raise ConnectionError(
                            "Subscription channel connection has been closed!"
                        )
                    await self.subscribe(request=msg.inner)
                    # Breaking out from the loop is required to regenerate
                    # component_data_receivers to be able to fulfil this
                    # subscription (later can be optimized by checking if
                    # an output channel already existed in the `subscribe()` method)
                    break
```

This is a real example from the SDK: https://github.com/frequenz-floss/frequenz-sdk-python/blob/1e01c2120cb96d703d3c3f6bb1fe1b738fa439d8/src/frequenz/sdk/data_ingestion/resampling/component_metrics_resampling_actor.py#L211-L240

This code is correct now, but there were 2 bugs that are very easy to add when handling a `read()` result that introduce data loss:

1. If the `break` is not in the handling of the last *event*, then *events* could be missed (and dropped by `ready()`.
2. If one uses `elif` instead of `if`, which is not intuitive when handling something that looks like a `switch`/`case`, we can again lose *events*.

The most common pattern to handle *event loops* (which `Select.ready()` looks a lot like) is to receive one event at the time. Also if we do this, we can enable the use of [structural pattern matching](https://peps.python.org/pep-0636/) via `match`.

### Proposed solution

Return only one ready *event* in `.ready()`.

### Use cases

I've seen these 2 mistakes made very often, and even found one in the channels documentation itself:

[...]

Issue #26, opened 2022-11-02
Allow sending `None`


leandro-lucarella-frequenz:
### What's needed?

Users should be able to send **any** value through a channel (or at least any serializable value if at some point channels can be over the wire). This includes `None`.

### Proposed solution

Allow sending `None` by raising an exception instead of returning `None` if a channel is closed.

### Alternatives and workarounds

Wrapping the `None` in another type (like `(None,)`).

shsms:
This would require internal representations to not use None to signal an empty channel.  A sentinal value could be used instead,  as suggested here: https://github.com/frequenz-floss/frequenz-channels-python/pull/42/files/206bbd7a056c32c452281a1bf390d12035f9f244#r1021864402

shsms:
Done in #83

PR #42, opened 2022-11-07
Avoid dropping of messages after breaking from `Select` blocks

shsms:
The current implementation of `Select.ready()`,  instead of just checking that some of the receivers have messages,  also fetches them.  

And in cases where there are multiple receivers with messages waiting to be read,  `Select.ready()` consumes the latest message from each of them,  and if subsequent user code decides to drop the `Select` object,  and recreates it for some reason,  the unprocessed messages that `Select.ready()` had fetched will get lost.  

This is incorrect usage of `Select`,  but the implementation should still try to ensure messages don't get dropped in such cases.  This PR addresses this issue as follows:

1. the `Receiver` interface now requires implementations to provide two private methods `_ready` and `_get`,  which are used to implement `__anext__` on the interface.  So `Receiver` objects would retain the original interface with `__anext__`, and `receive`,  but now provide methods to only check if a message is ready or wait until one is,  and a separate method to get a message after it becomes ready.
2. Update `Select.ready()` to use the `Receiver._ready()` method to know when a message is ready and fetch the message with `_get()` only when the user code is trying to fetch the message.  
3. This would mean the `Select` implementation no longer be able to operate on async iterators like it used to,  but that's probably a good thing,  taking away some of its power.


leandro-lucarella-frequenz:
Have you seen https://github.com/frequenz-floss/frequenz-channels-python/issues/27#issuecomment-1303299010? I'm sorry to be **that guy**, but I'm getting the impression we are making things more complicated than they should be. We don't really need to know when stuff is ready, we just need to take the first consumable message and `yield` it.

Am I missing something? Is there any use case you see that can't be done by making `select()` an async iterator itself instead of the current approach?

I know it is a big departure from the current interface, but also dropping support for async iterators feels going backwards, IHMO we should point to be more compatible and idiomatic about async iterators.

shsms:
> We don't really need to know when stuff is ready, we just need to take the first consumable message and yield it.

how would we implement that?  We already call `asyncio.wait` with `return_when=asyncio.FIRST_COMPLETED`.  It still returns multiple values if multiple have values: https://github.com/frequenz-floss/frequenz-channels-python/blob/v0.x.x/src/frequenz/channels/select.py#L119-L121

 I looked at the other comment,  but I think that would have the same problem.  It looks like a nice interface and i would like to migrate to it.

Maybe you have a design where messages won't get lost and I'm just missing it, but the way I see it, we'll still have to solve the problem of losing messages if users use 'break', even if we switch to the other design.

leandro-lucarella-frequenz:
> how would we implement that? We already call `asyncio.wait` with `return_when=asyncio.FIRST_COMPLETED`. It still returns multiple values if multiple have values: [https://github.com/frequenz-floss/frequenz-channels-python/blob/v0.x.x/src/frequenz/channels/select.py#L119-L121](https://github.com/frequenz-floss/frequenz-channels-python/blob/v0.x.x/src/frequenz/channels/select.py?rgh-link-date=2022-11-10T13%3A24%3A00Z#L119-L121)

I was thinking about just `yield`ing one result at the time when doing the `for item in done:` loop. But you are right, if a user `break`s the `async for msg in select(...):` then we still lose messages :(

I can't believe there is no option for `async.wait()` that only returns the first done item! That's the root of the problem here.

Looking at the docs it looks like [`as_completed()`](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed) is what we want:

```py
import asyncio

async def f() -> int:
    return 1

async def main():
    a1 = asyncio.create_task(f(), name='a1')
    a2 = asyncio.create_task(f(), name='a2')
    done, pending = await asyncio.wait((a1, a2))
    print(f"wait: {done=} {pending=}")
    done = [i.result() for i in done]
    print(f"{done}")

    a1 = asyncio.create_task(f(), name='a1')
    a2 = asyncio.create_task(f(), name='a2')
    print("as_completed:")
    for coro in asyncio.as_completed((a1, a2)):
        print(await coro)

asyncio.run(main())
```

```
wait: done={<Task finished name='a1' coro=<f() done, defined at wait.py:3> result=1>, <Task finished name='a2' coro=<f() done, defined at wait.py:3> result=1>} pending=set()
[1, 1]
as_completed:
1
1
```

sahas-subramanian-frequenz:
> I can't believe there is no option for async.wait() that only returns the first done item! That's the root of the problem here.

I think that makes sense.  We make calls to `__anext__` or `receive`,  and when those functions complete,  a value has already been pulled out of the receiver,  and is sitting in the `future`.  When we `break` from the loop,  the `wait` task and the ready `future`s get dropped,  and only the receivers remain.  But the values have already gone out of the receivers.

The only solution is for the receivers to be able to inform downstream when a value is ready,  without having to give out the value,  and downstream can fetch the value when it is ready.

And it doesn't really complicate the implementation,  or the external API.  It is still clearly defined,  and even if external parties try to make their own channel implementations,  it shouldn't pose a problem.

> Looking at the docs it looks like [as_completed()](https://docs.python.org/3/library/asyncio-task.html#asyncio.as_completed) is what we want:

I'm pretty sure `as_completed` has the same problem as well,  because it is running calls to `receive` concurrently,  so multiple values are going to get pulled out of receivers,  but we are processing them only one at a time.

leandro-lucarella-frequenz:
> I'm pretty sure `as_completed` has the same problem as well, because it is running calls to `receive` concurrently, so multiple values are going to get pulled out of receivers, but we are processing them only one at a time.

No it doesn't:

```py
import asyncio

async def f() -> int:
    return 1

async def main():
    a1 = asyncio.create_task(f(), name='a1')
    a2 = asyncio.create_task(f(), name='a2')
    done, pending = await asyncio.wait((a1, a2))
    print(f"wait: {done=} {pending=}")
    done = [i.result() for i in done]
    print(f"{done}")

    a1 = asyncio.create_task(f(), name='a1')
    a2 = asyncio.create_task(f(), name='a2')
    print("as_completed:")
    for coro in asyncio.as_completed((a1, a2)):
        print(await coro)
        break
    print("as_completed2:")
    for coro in asyncio.as_completed((a1, a2)):
        print(await coro)
        break

asyncio.run(main())
```

```
wait: done={<Task finished name='a1' coro=<f() done, defined at wait.py:3> result=1>, <Task finished name='a2' coro=<f() done, defined at wait.py:3> result=1>} pending=set()
[1, 1]
as_completed:
1
as_completed2:
1
```

I still didn't test how it behaves if some tasks are not ready yet (the docs are really bad, it says very little, and the name of the function doesn't help either).

leandro-lucarella-frequenz:
But OK, I have been trying some thing around and I think I'm not always taking into account the difference between a task and an async iterator.

But while paying with the options I notice that if we do an `async.wait()` over an already completed task, it will return immediately with the completed task. The received value is stored in the task, so we can retrieve it again. So I think the trick might rely on simply not removing the `done` tasks from our internal pending list until it was `yield`ed.

I will play a little bit more around with that concept before keep spamming this PR, sorry about that, I should have probably started a separate discussion about it. :grimacing:

leandro-lucarella-frequenz:
> The received value is stored in the task, so we can retrieve it again. So I think the trick might rely on simply not removing the `done` tasks from our internal pending list until it was `yield`ed.

But if we break we lose the status of the loop, so we lose messages, so yeah, this doesn't work. I have the feeling there should be a way to do it, but I should really stop spamming this PR until I have a proven working solution :grimacing:

sahas-subramanian-frequenz:
``` python
    a1 = asyncio.create_task(f(), name='a1')
    a2 = asyncio.create_task(f(), name='a2')
    print("as_completed:")
    for coro in asyncio.as_completed((a1, a2)):
        print(await coro)
        break
    print("as_completed2:")
    for coro in asyncio.as_completed((a1, a2)):
        print(await coro)
        break
```

yes,  but you are operating on the same tasks `a1` and `a2` after you break out of the loop.  But in the select implementation,  `a1` and `a2` will be owned by `select`,  and when we break out of the loop,  the task objects get lost and hence the values they hold.

leandro-lucarella-frequenz:
OK, I think I can conclude that `as_completed()` does exactly what we want, **except** it doesn't return the finished task but its `.result()`, so we don't have a way to know which of all the tasks finished, so we can't re-schedule it (unless we start wrapping the tasks to return also some way to identify which one finished as part of the `.result()`).

Another option would be to implement our own `as_completed()`, the code is not that complicated, which returns the task [instead of the `.result()`](https://github.com/python/cpython/blob/db115682bd639a2642c617f0b7d5b30cd7d7f472/Lib/asyncio/tasks.py#L608): https://github.com/python/cpython/blob/db115682bd639a2642c617f0b7d5b30cd7d7f472/Lib/asyncio/tasks.py#L561-L615

This approach is not making the code simpler, but an advantage over this approach, IMHO, is that we keep the complexity in `select()` only, and receivers can still be simple async iterators, which for me it makes a lot of sense, and avoid people wanting to provide new channels to worry about implementing readiness primitives.

I gave it a shot by doing the mapping of the done coroutine returned by `as_completed()` and our pending items, and the code is greatly simplified but for some reason it is not working completely as expected.

https://github.com/frequenz-floss/frequenz-channels-python/pull/46

leandro-lucarella-frequenz:
> yes, but you are operating on the same tasks `a1` and `a2` after you break out of the loop. But in the select implementation, `a1` and `a2` will be owned by `select`, and when we break out of the loop, the task objects get lost and hence the values they hold.

Not if we break the loop but keep the `Select` object alive. But yeah, I see your points, that would lead to a `select()` function implementation that it is not correct. I see now we have another layer of complexity, which is we are consuming from the async iterators in a function running inside a task, so if the task already consumed the value and we cancel the task for some reason, then the value will be lost too. The alternative I presented above suffers from this issue too.

So I think you finally convinced me that we can't have both a nice simple **and safe** async iterator interface and not introduce the *ready test* to receivers.

I really hate moving the complexity to channels implementators, but I guess there is no way around that is not moving complexity to users of `select()`, which is even worse.

src/frequenz/channels/anycast.py
```diff
@@ -162,23 +162,38 @@ def __init__(self, chan: Anycast[T]) -> None:
...
+    def _get(self) -> T:
+        """Return the latest value once `_ready()` is complete.
+
+        Raises:
+            EOFError: When called before a call to `_ready()` finishes.
+
+        Returns:
+            The next value that was received.
+        """
+        if self._next is None:
```
leandro-lucarella-frequenz: Now that values passed via channels could be `None`, I think using `None` here could be problematic. You could use a [sentinel value](https://peps.python.org/pep-0661/) to indicate there is no `next`.

src/frequenz/channels/anycast.py
```diff
@@ -162,23 +162,38 @@ def __init__(self, chan: Anycast[T]) -> None:
...
+            raise EOFError("_get was called before a call to _ready finished.")
```
leandro-lucarella-frequenz: I would `assert` on this or use some other special error `ReadyNotCalledError` or something, as this doesn't really have anything to do with EOF. It's just a programming error, you didn't call the methods in the order you were supposed to.

src/frequenz/channels/base_classes.py
```diff
@@ -31,12 +40,40 @@ async def send(self, msg: T) -> bool:
...
+
+        `_ready()` must be called before each call to `_get()`.
 
         Returns:
-            `None`, if the channel is closed, a message otherwise.
+            The next value received.
+
+        Raises:
+            StopAsyncIteration: if the underlying channel is closed.
         """
```
leandro-lucarella-frequenz: I think I would make these public. Any reason for them to be private? Even if I would consider advanced usage to access these, I don't see a reason not to make them part of the public interface if they will be such a fundamental construct for receivers, and other people might want to build stuff on top of them.

src/frequenz/channels/base_classes.py
```diff
@@ -12,6 +12,15 @@
 U = TypeVar("U")
 
 
+class ChannelClosedError(Exception):
```
leandro-lucarella-frequenz: I would inherit from `RuntimeError` and probably also add a base exception class for all errors generated by this library. In my attempt to do this I added a file `exceptions.py` (I guess it can also go to `base_classes`) with:

```py
"""
Exception classes for channels.

Copyright
Copyright © 2022 Frequenz Energy-as-a-Service GmbH

License
MIT
"""


from typing import Any


class ChannelError(RuntimeError):
    """Base channel error.

    All exceptions generated by channels inherit from this exception.
    """

    def __init__(self, message: Any, channel: Any = None):
        super().__init__(message)
        self.channel: Any = channel


class ChannelClosedError(ChannelError):
    """Error raised when trying to operate on a closed channel."""

    def __init__(self, channel: Any = None):
        super().__init__(f"Channel {channel} was closed", channel)
```

src/frequenz/channels/base_classes.py
```diff
@@ -31,12 +40,40 @@ async def send(self, msg: T) -> bool:
...
-    async def receive(self) -> Optional[T]:
-        """Receive a message from the channel.
+    async def _ready(self) -> None:
+        """Wait until the receiver is ready with a value.
+
+        Once a call to `_ready` has finished, the value should be read with a call to
+        `_get()`.
+
+        Raises:
+            StopAsyncIteration: if the underlying channel is closed.
```
leandro-lucarella-frequenz: Why not raising `ChannelClosedError` in `ready()`? I just just out of convenience because it is used in `__anext__()` but I think from the interface POV it makes more sense to to `raise` `ChannelClosedError` here.

src/frequenz/channels/anycast.py
```diff
@@ -162,23 +162,38 @@ def __init__(self, chan: Anycast[T]) -> None:
...
+                raise StopAsyncIteration()
             async with self._chan.recv_cv:
                 await self._chan.recv_cv.wait()
-        ret = self._chan.deque.popleft()
+        self._next = self._chan.deque.popleft()
         async with self._chan.send_cv:
             self._chan.send_cv.notify(1)
-        return ret
+
+    def _get(self) -> T:
```
leandro-lucarella-frequenz: I would name this something more like `consume()` or something else that indicates more clearly that it is a *destructive* operation. For me `get()` sounds too much as if it were idempotent and you could call `get()` many times and always get the same.

I think it is important to also document more clearly that users are expected to use both `ready()` and `consume()` (or whatever is called) together in both functions documentation.

(off-topic note: I would like to rename also the `get_receiver/sender()` methods from channels to something more like `new_receiver/sender()`, this also has already cause confusion and people expected to receive always the same receiver when calling `get_receiver()`)

benchmarks/benchmark_anycast.py
```diff
@@ -51,10 +51,7 @@ async def benchmark_anycast(
     recv_trackers = [0]
 
     async def update_tracker_on_receive(chan: Receiver[int]) -> None:
-        while True:
-            msg = await chan.receive()
-            if msg is None:
-                return
+        async for _ in chan:
```
leandro-lucarella-frequenz, line 54: :heart:

src/frequenz/channels/broadcast.py
```diff
@@ -257,31 +257,32 @@ def __len__(self) -> int:
...
-        `EOFError`.
+    async def _ready(self) -> None:
+        """Wait until the receiver is ready with a value.
 
         Raises:
-            EOFError: when the receiver has been converted into a `Peekable`.
-
-        Returns:
-            `None`, if the channel is closed, a message otherwise.
+            EOFError: When called before a call to `_ready()` finishes.
```
leandro-lucarella-frequenz: This doesn't look right, this is `ready()` itself! :)

src/frequenz/channels/broadcast.py
```diff
@@ -257,31 +257,32 @@ def __len__(self) -> int:
...
+            StopAsyncIteration: if the underlying channel is closed.
         """
         if not self._active:
             raise EOFError("This receiver is no longer active.")
```
leandro-lucarella-frequenz, line 260: While we are introducing our own exceptions I would also consider adding a specific exception for this (maybe `ReceiverInactiveError`?), `EOFError` is only used by `input()`, so it feels a bit weird to raise this exception here.

src/frequenz/channels/broadcast.py
```diff
@@ -257,31 +257,32 @@ def __len__(self) -> int:
...
+                raise StopAsyncIteration()
             async with self._chan.recv_cv:
                 await self._chan.recv_cv.wait()
+
+    def _get(self) -> T:
+        """Return the latest value once `_ready` is complete.
+
+        Returns:
+            The next value that was received.
+        """
```
leandro-lucarella-frequenz, line 277: IMHO this confirms that in the interface `get()` (or `consume()` ;) ) shouldn't raise any specific exception if it is misused (called before `ready()`), as here we can't even have an `assert`, if it is misused, it will misbehave, so I would go with `assert` in `Anycast`.

src/frequenz/channels/select.py
```diff
@@ -121,30 +148,29 @@ async def ready(self) -> bool:
...
             else:
-                result = item.result()
+                result = recv
             self._ready_count += 1
-            self._result[name] = _Selected(result)
-            # if channel or AsyncIterator is closed
+            self._result[name] = _ReadyReceiver(result)
+            # if channel or Receiver is closed
             # don't add a task for it again.
             if result is None:
```
leandro-lucarella-frequenz, line 158: This also probably needs to use a sentinel now that `None` is a valid possible result (and add some test to make sure `None` can be passed via channels).

src/frequenz/channels/utils/file_watcher.py
```diff
@@ -52,6 +53,7 @@ def __init__(
                 and pathlib.Path(path_str).is_file()
             ),
         )
+        self._changes: Set[FileChange] = set()
```
leandro-lucarella-frequenz, line 56: It looks like ordering is important here and I don't think `set` preserves order. Also, why do we want to compress `FileChange`s that are the same, suppose we have this sequence:

1. `/a` is added
2. `/a` is deleted
3. `/a` is added again

With this implementation we could only get one add and one deletion so we would think the file doesn't exist but it does.

If this introduced a bug that wasn't detected by test, we should probably add a test for this too :)

src/frequenz/channels/utils/file_watcher.py
```diff
@@ -62,19 +64,26 @@ def __del__(self) -> None:
...
+        self._changes = await self._awatch.__anext__()
+
+    def _get(self) -> pathlib.Path:
+        change = self._changes.pop()
+        # Tuple of (Change, path) returned by watchfiles
+        if change is None or len(change) != 2:
+            raise StopAsyncIteration()
+        _, path_str = change
+        path = pathlib.Path(path_str)
+        return path
```
leandro-lucarella-frequenz, line 98: Oh, we are not even reporting what kind of change was observed? :thinking: :scream:

leandro-lucarella-frequenz (commented):
This looks good. I like it. It's also not that bad in terms of complexity as it only splits the work that needed to be done by `receive()`, so I'm less worried about complexity for channels implementors too.

Before this PR we also had the issue about potential messages being lost when `break`ing in `Merge`/`MergeNamed`, right?

This also needs a `RELEASE_NOTES` entry!

In a followup PR I think we should definitely go for the async iterable interface for `Select.ready()`, it will simplify the implementation too (as we won't need the extra `class` for lazy evaluation for example).

src/frequenz/channels/anycast.py
shsms: Will do this as part of #26,  including tests, etc.

src/frequenz/channels/anycast.py
shsms: updated to `ready` and `consume`

src/frequenz/channels/anycast.py
shsms: switched to `assert`

src/frequenz/channels/base_classes.py
shsms: done

src/frequenz/channels/base_classes.py
shsms: true,  done.

src/frequenz/channels/broadcast.py
shsms: done

src/frequenz/channels/utils/file_watcher.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-channels-python/issues/51

src/frequenz/channels/utils/file_watcher.py
shsms: added to #51

src/frequenz/channels/select.py
shsms: Doing as part of #26

src/frequenz/channels/base_classes.py
shsms: done

src/frequenz/channels/broadcast.py
shsms: made an issue: https://github.com/frequenz-floss/frequenz-channels-python/issues/52

src/frequenz/channels/utils/file_watcher.py
leandro-lucarella-frequenz: OK, this turned out to be invalid, because `awatch` returns a `set`, so there is no order anyway. It is kind of weird that you could potentially get unordered changes on the same file in the set. Maybe the underlying library only accumulates one change per file in every loop, but I couldn't find any mention of this neither in the Python or the Rust projects.

src/frequenz/channels/utils/file_watcher.py
leandro-lucarella-frequenz: I created #53 for this.

src/frequenz/channels/utils/file_watcher.py
```diff
@@ -62,19 +64,35 @@ def __del__(self) -> None:
...
+            ChannelClosedError: When the channel is closed.
+
+        Returns:
+            The next change that was received.
+        """
+        assert self._changes, "calls to `consume()` must be follow a call to `ready()`"
+        change = self._changes.pop()
+        # Tuple of (Change, path) returned by watchfiles
+        if change is None or len(change) != 2:
+            raise ChannelClosedError()
```
leandro-lucarella-frequenz, line 95: This is also weird, but pre-existing, so I created https://github.com/frequenz-floss/frequenz-channels-python/issues/54.

src/frequenz/channels/base_classes.py
leandro-lucarella-frequenz: The documentation is still outdated here.

src/frequenz/channels/broadcast.py
```diff
@@ -249,31 +249,33 @@ def __len__(self) -> int:
...
-        `EOFError`.
+    async def ready(self) -> None:
+        """Wait until the receiver is ready with a value.
 
         Raises:
-            EOFError: when the receiver has been converted into a `Peekable`.
-
-        Returns:
-            `None`, if the channel is closed, a message otherwise.
+            EOFError: When called before a call to `ready()` finishes.
```
leandro-lucarella-frequenz: Super minimal, but this still needs updating (`ready()` can't be called before `ready()` :D).

leandro-lucarella-frequenz:
Needs a couple of documentation fixes and it is ready to go! Feel free to merge after those fixes.

src/frequenz/channels/broadcast.py
shsms: oh, wow. :D  fixed.

src/frequenz/channels/base_classes.py
shsms: fixed.

sahas-subramanian-frequenz (approved):

PR #45, opened 2022-11-11
Ensure deleted receivers get cleaned up

### LINKED ISSUE #44: `Receiver`s are not getting deleted when they go out of scope
### What happened?

`Receiver`s are not getting destroyed as soon as they go out of scope,  because channel implementations hold on to a reference to the `Receiver`s,  so that they can put new messages into them.

So when a the task that created a receiver dies, the receiver doesn't, but nobody is reading from it,  and the logs get flooded with

```
Broadcast receiver [chan_name] is full. Oldest message was dropped.
```

`Receiver`s that go out of scope should get cleaned up.

### Affected version(s)

_No response_

### Affected part(s)

Channels, `Broadcast`, `Bidirectional`, etc. (part:channels)

### Extra information

The solution appears to be to use a [weakref](https://docs.python.org/3/library/weakref.html) to hold on to the receivers in the channel implementations.

And the `send` methods should check and cleanup any weakrefs to destroyed receivers.

shsms:
`Broadcast` receivers were not getting cleaned up when then go out of scope,  because `Broadcast` instances were holding on to a strong reference to them.  And because these receivers were not being read from anymore,  but were still getting messages,  their buffers would overflow,  and the logs would get flooded with warnings.

This PR fixes this issue by storing just weak references to the receivers in the `Broadcast` instances.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -20,9 +20,11 @@ time.
 
 ## Bug Fixes
 
+* [`Broadcast`] receivers now get cleaned up once they go out of scope.
+
 * [`Timer`] now returns [timezone-aware] `datetime` objects using UTC as
   timezone.
 
-
+[`Broadcast`]: https://frequenz-floss.github.io/frequenz-channels-python/v0.11/reference/frequenz/channels/#frequenz.channels.Broadcast
 [`Timer`]: https://frequenz-floss.github.io/frequenz-channels-python/v0.11/reference/frequenz/channels/#frequenz.channels.Timer
 [timezone-aware]: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects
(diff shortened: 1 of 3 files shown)


leandro-lucarella-frequenz (dismissed):
LGTM, but I would add a note on the `RELEASE_NOTES` as it seems like an important bug fix, and also I would add a better explanation of what kind of problems we observed before this fix in the PR description (which will be used as the merge commit message).

shsms:
> LGTM, but I would add a note on the `RELEASE_NOTES` as it seems like an important bug fix, and also I would add a better explanation of what kind of problems we observed before this fix in the PR description (which will be used as the merge commit message).

Both are done.

sahas-subramanian-frequenz (approved):

leandro-lucarella-frequenz:
Cheater! :laughing: 

image

PR #46, opened 2022-11-14
PoC: Make ready() an async iterable

leandro-lucarella-frequenz:
Alternative approach to #42.

### DIFF
--- a/tests/test_select.py
+++ b/tests/test_select.py
@@ -45,25 +45,37 @@ async def test_select() -> None:
         ch1: Sender[int], ch2: Sender[int], ch3: Sender[int], queue: AsyncIterable
     ) -> None:
         for ctr in range(5):
-            await ch1.send(ctr + 1)
-            await ch2.send(ctr + 101)
-            await ch3.send(ctr + 201)
-            await queue.add(ctr + 301)
+            print(f"SEND: ch1: {ctr+101}")
+            await ch1.send(ctr + 101)
+            print(f"SEND: ch2: {ctr+201}")
+            await ch2.send(ctr + 201)
+            print(f"SEND: ch3: {ctr+301}")
+            await ch3.send(ctr + 301)
+            print(f"SEND: ch4: {ctr+401}")
+            await queue.add(ctr + 401)
+        print(f"CLOSE: ch1")
         await chan1.close()
+        print(f"SEND: ch2: 1000")
         await ch2.send(1000)
+        print(f"CLOSE: ch2")
         await chan2.close()
+        print(f"CLOSE: ch3")
+        print(f"DONE: ch4")
         await chan3.close()
         queue.done = True
 
     queue = AsyncIterable()
+    ch1 = chan1.get_receiver()
+    ch2 = chan2.get_receiver()
+    ch3 = chan3.get_receiver()
 
     senders = asyncio.create_task(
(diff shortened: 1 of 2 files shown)


leandro-lucarella-frequenz:
The `test_select.py` test fails with:

```
>       assert results == expected_results
E       assert [101, 102, 103, 104, 105, -1, 201, 401, 202, 402, 203, 403, 204, 404, 205, 405, 1000, -4, -2] == [101, 201, 401, 102, 202, 402, 103, 203, 403, 104, 204, 404, 105, 205, 405, -1, 1000, -4, -2]
E         At index 1 diff: 102 != 201
E         Full diff:
E           [
E            101,
E         +  102,
E         +  103,
E         +  104,
E         +  105,
E         +  -1,
E            201,
E            401,
E         -  102,
E            202,
E            402,
E         -  103,
E            203,
E            403,
E         -  104,
E            204,
E            404,
E         -  105,
E            205,
E            405,
E         -  -1,
E            1000,
E            -4,
E            -2,
E           ]
```

For some reason is treating `ch1` differently. I could understand if somehow this way the order is not preserved, guaranteed, but I don't understand why the special behavior only with `ch1`.

tests/test_select.py
```diff
@@ -45,75 +45,82 @@ async def send(
         ch1: Sender[int], ch2: Sender[int], ch3: Sender[int], queue: AsyncIterable
     ) -> None:
         for ctr in range(5):
-            await ch1.send(ctr + 1)
-            await ch2.send(ctr + 101)
-            await ch3.send(ctr + 201)
-            await queue.add(ctr + 301)
+            print(f"SEND: ch1: {ctr+101}")
```
leandro-lucarella-frequenz, line 48: Also originally, before I introduced these `print()`s, I got the processing of `ch1` at the end, not the beginning:

```
>       assert results == expected_results
E       assert [101, 201, 401, 202, 402, 203, 403, 204, 404, 205, 405, 1000, -4, -2, 102, 103, 104, 105, -1] == [101, 201, 401, 102, 202, 402, 103, 203, 403, 104, 204, 404, 105, 205, 405, -1, 1000, -4, -2]
E         At index 3 diff: 202 != 102
E         Full diff:
E           [
E            101,
E            201,
E            401,
E         -  102,
E            202,
E            402,
E         -  103,
E            203,
E            403,
E         -  104,
E            204,
E            404,
E         -  105,
E            205,
E            405,
E         -  -1,
E            1000,
E            -4,
E            -2,
E         +  102,
E         +  103,
E         +  104,
E         +  105,
E         +  -1,
E           ]
```

This is reproduceable for me, if I remove the prints I still get this.

leandro-lucarella-frequenz:
Conceptually this is what I got in mind, the code looks much much simpler, too bad it's not really working for some reason.

@sahas-subramanian-frequenz since you have been more involved using tasks maybe you get some idea of why this could be happening?

As an extra clarification, I had to use `next(as_completed(...),...)` instead of just a for loop because of course we need to re-create the task once it finished.

leandro-lucarella-frequenz:
I will close this as it doesn't really work because tasks are still consuming from the underlaying async iterators, so if the loop is broken and the `Select` object is destroyed, data will still be lost.

But we should definitely go for the async iterable interface for ready. I created #47 for this.

PR #48, opened 2022-11-16
Use "aware" datetimes

leandro-lucarella-frequenz:
When creating datetimes we need to make sure to include timezone information, otherwise we will get exceptions raised when comparing to other "aware" datetime objects. Using "native" datetime object (without timezone information) is very dangerous as we might mix datetimes created for different timezones without knowing, introducing very obscure bugs.

More info about "aware" and "native" datetimes: https://docs.python.org/3/library/datetime.html#aware-and-naive-objects

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,12 @@
 
 ## Summary
 
-<!-- Here goes a general summary of what this release is about -->
+The project has a new home!
+
+https://frequenz-floss.github.io/frequenz-channels-python/
+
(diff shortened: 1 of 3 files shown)


leandro-lucarella-frequenz:
For more context, this is affecting some tests in the SDK, which was really hard to figure out.

sahas-subramanian-frequenz (approved):
Oops,  that makes sense.  Thanks Luca!

leandro-lucarella-frequenz:
It is quite *inconvenient* (:hankey:) that Python doesn't use "aware" `datetime`s by default.

PR #55, opened 2022-11-22
Clean up public API

### LINKED ISSUE #50: Make the public interface of the library more clean
### What's needed?

We need to have a cleaner public interface in terms of modules and where each class is exposed. Right now we have everything exposed in some internal modules (like `frequenz.channels.broadcast.Broadcast`), but what it is really meant to be used is also exposed via `frequenz.channels` directly (for example `frequenz.channels.Broadcast`).

This ends up in a lot of duplication and a lot of *irrelevant* symbols being shown in the API documentation (https://frequenz-floss.github.io/frequenz-channels-python/next/reference/frequenz/channels/).

### Proposed solution

Hide all internal modules by prepending a `_` to the module name.

I would also expose anything that is not a channel in a `util` package. So:

* `frequenz.channels.`
  * `Anycast`
  * `Bidirectional.Handle` (declare `Handle` inside `Bidirectional`)
  * `Broadcast`
  * `Peekable`
  * `Receiver`
  * `Sender`

* `frequenz.channels.util.`
  * `FileWatcher`
  * `Merge`
  * `MergeNamed`
  * `Select`
  * `Timer`

* Don't expose `BufferedReceiver` anymore, and maybe move it to `Broadcast`, as it doesn't look like it's being use as an interface.

### Use cases

_No response_

The obvious alternative is to keep all symbols exposed directly from `frequenz.channels`.

### Discussed in https://github.com/frequenz-floss/frequenz-channels-python/discussions/30

<div type='discussions-op-text'>

<sup>Originally posted by **leandro-lucarella-frequenz** November  3, 2022</sup>
Now that we have API documentation, we are exposing every module in the documentation, when the actual classes that are intended to be used by the users are all imported publicly in `frequenz.channels`.

For a library this small I think it makes sense con only consider that module our public API, this will also allow us to restructure the project in a non-breaking way.

The de facto way to mark modules and packages *internal* is to prefix a `_`, so all modules and packages will have to be prefixed this way if we want to go this route.</div>

leandro-lucarella-frequenz:
This PR hides all internal modules by prepending a `_` to the module name.

It also expose anything that is not a channel (or core channel classes) in
a `util` package. So this is how classes are exposed:

* `frequenz.channels.Anycast`
* `frequenz.channels.Broadcast`
* `frequenz.channels.Anycast`
* `frequenz.channels.Bidirectional`
* `frequenz.channels.Broadcast`
* `frequenz.channels.Peekable`
* `frequenz.channels.Receiver`
* `frequenz.channels.Sender`
* `frequenz.channels.util.Merge`
* `frequenz.channels.util.MergeNamed`
* `frequenz.channels.util.FileWatcher`
* `frequenz.channels.util.Select`
* `frequenz.channels.util.Timer`

This also completely removes `BufferedReceiver` and moves `BidirectionalHandle`
to `Bidirectional.Handle` and `EventType` to `FileWatcher.EventType`.

While at it, add also a short summary of each module contents.


leandro-lucarella-frequenz:
Adding high priority so we can release and update the SDK soon.

leandro-lucarella-frequenz:
The resulting documentation can be seen here: https://leandro-lucarella-frequenz.github.io/frequenz-channels-python/next/reference/frequenz/channels/

shsms (approved):

sahas-subramanian-frequenz (approved):
Sorry I keep approving with the wrong account.

PR #61, opened 2022-12-30
Improve and add exceptions

### LINKED ISSUE #52: Make a `ReceiverInactiveError` exception for receivers turned to peekables
While we are introducing our own exceptions I would also consider adding a specific exception for this (maybe `ReceiverInactiveError`?), `EOFError` is only used by `input()`, so it feels a bit weird to raise this exception here.

_Originally posted by @leandro-lucarella-frequenz in https://github.com/frequenz-floss/frequenz-channels-python/pull/42#discussion_r1021982329_

leandro-lucarella-frequenz:
- Properly await for service task
- Make output more verbose for darglint
- Add base exception
- Make `send()` raise a `SenderError`
- Make receivers raise `ReceiverError`
- Move exceptions into its own module
- Add a `ReceiverInvalidatedError`
- Don't raise exceptions in Receiver.ready()

Fixes #52.


leandro-lucarella-frequenz:
Updated to the latest head and added a commit that moves exception raising from `ready()` to `consume()` (previously in #62 only).

ping @sahas-subramanian-frequenz

src/frequenz/channels/_bidirectional.py
```diff
@@ -23,26 +23,48 @@ class Handle(Sender[V], Receiver[W]):
...
+                # ChannelError having the information about the Bidirectional
+                # channel to hide (at least partially) the underlaying
+                # Broadcast channels we use.
+                if isinstance(err.__cause__, ChannelError):
+                    this_chan_error = ChannelError(
+                        f"Error in the underlying channel {err.__cause__.channel}: {err.__cause__}",
+                        self._chan,  # pylint: disable=protected-access
+                    )
+                    this_chan_error.__cause__ = err.__cause__
+                    err.__cause__ = this_chan_error
```
shsms, line 68: don't you have to re-raise the exception here because you're catching it?

src/frequenz/channels/_bidirectional.py
```diff
@@ -67,8 +75,26 @@ async def send(self, msg: V) -> None:
...
```
shsms: missing a raise here too?

src/frequenz/channels/_base_classes.py
```diff
@@ -133,6 +40,11 @@ async def __anext__(self) -> T:
             StopAsyncIteration: if the receiver stopped producing messages.
             ReceiverError: if there is some problem with the receiver.
         """
+        # Break cyclic import
+        from ._exceptions import (  # pylint: disable=import-outside-toplevel,cyclic-import
+            ReceiverStoppedError,
+        )
```
shsms: does this have a perf cost,  importing from inside `__anext__`?

It looks like the cyclic importing is necessary just because of type hint dependencies in `_exceptions.py`.  It might be interesting to examine switching to automatically generated typing stubs,  then types become independent:  https://mypy.readthedocs.io/en/stable/stubgen.html

It is hard for me to see the benefit in pulling the exceptions into a separate package,  when we're having to work around ourselves when importing the package.

src/frequenz/channels/_bidirectional.py
leandro-lucarella-frequenz: YUP! Good... *catch* :clown_face:

src/frequenz/channels/_bidirectional.py
leandro-lucarella-frequenz: Yes, but that was fixed when moving the exceptions to `consume()` at least. I'll amend the original location to pretend I never make mistakes anyway :P

src/frequenz/channels/_base_classes.py
leandro-lucarella-frequenz: > does this have a perf cost

Good question, my guess is it shouldn't be too much, as import should first check if the module was already imported, so it should be just an extra lookup in the modules dict, which compared to sync primitives should be negligible, but I agree is probably the worse place to break a circular dependency.

I don't think generating stubs is a good idea, it was more of a hack for having at least *some* type info for libs that didn't use native type hints, but if you look at the stubgen example in the link you sent, it doesn't even infer the type of the method arguments.

> Stubgen generates draft stubs. The auto-generated stub files often require some manual updates, and most types will default to `Any`.

> It is hard for me to see the benefit in pulling the exceptions into a separate package, when we're having to work around ourselves when importing the package.

Yeah, for me is just an organizational improvement, is easier to find the code and to avoid noise when looking at unrelated stuff, but I do agree that it might not be the best approach if it needs *workarounding*.

I'll see if there is a better way to avoid the dependency loops and if not I might just move the exceptions back to the base classes module.

leandro-lucarella-frequenz: OK, I found an *elegant* (by Python type hints standards :trollface:) way to break the circular dependency: [`typing.TYPE_CHECKING`](https://docs.python.org/3/library/typing.html#typing.TYPE_CHECKING).

leandro-lucarella-frequenz:
[Updated](https://github.com/frequenz-floss/frequenz-channels-python/compare/c390cf27c0497cecde35d21a9de1241927da5c26..c463354124a099a99493155e7ad53f9986b9aa90):

* Remove local import to break circular dependency and use `TYPE_CHECKING` instead.
* Add the missing re-raise statement in `Bidirectional`
* Properly `await` for the service task in the tests to avoid `pytest` errors in `Bidirectional` tests
* Add tests to check exceptions are properly raised in `Bidirectional` (and they are also chained properly)

sahas-subramanian-frequenz (approved):

PR #63, opened 2023-01-23
Add method to stop Select

ela-kotulska-frequenz:
Add method to stop `Select` instance and cleanup any
pending task.
Select stores async tasks that listen on the receivers.
If select is removed without stopping the receivers tasks,
then we get error: Task was destroyed but it is pending!

We should stop all internal tasks before destoring Select.
Destructor can't be async, so we do that with extra method `stop`.
This method should be called when Select is no longer needed.

Signed-off-by: ela-kotulska-frequenz <elzbieta.kotulska@frequenz.com>

### DIFF
--- a/src/frequenz/channels/util/_select.py
+++ b/src/frequenz/channels/util/_select.py
@@ -63,6 +63,9 @@ class Select:
 
     [Receiver][frequenz.channels.Receiver]s also function as `Receiver`.
 
+    When Select is no longer needed, then it should be stopped using
+    `self.stop()` method. This would cleanup any internal pending async tasks.
+
     Example:
         For example, if there are two receivers that you want to
         simultaneously wait on, this can be done with:
@@ -105,6 +108,13 @@ class Select:
         """Cleanup any pending tasks."""
         for task in self._pending:
             task.cancel()
+
+    async def stop(self) -> None:
+        """Stop the `Select` instance and cleanup any pending tasks."""
+        for task in self._pending:
+            task.cancel()
+        await asyncio.gather(*self._pending, return_exceptions=True)
+        self._pending = set()
 
     async def ready(self) -> bool:
         """Wait until there is a message in any of the receivers.


src/frequenz/channels/util/_select.py
```diff
@@ -106,6 +106,13 @@ def __del__(self) -> None:
         for task in self._pending:
             task.cancel()
 
+    async def stop(self) -> None:
+        """Cleanup any pending tasks."""
```
shsms:
```suggestion
        """Stop the `Select` instance and cleanup any pending tasks.
        
        Once `stop` is called, the `Select` instance can no longer be used.
        """
```

ela-kotulska-frequenz: Done

src/frequenz/channels/util/_select.py
```diff
@@ -63,6 +63,9 @@ class Select:
 
     [Receiver][frequenz.channels.Receiver]s also function as `Receiver`.
 
+    When Select is no longer needed, then it should be stopped using
+    `self.stop()` method. This would stop all stored receivers, too.
```
shsms: it wouldn't stop the receivers,  the receivers will still be usable if people hold on to references to the receivers.

ela-kotulska-frequenz: Done

sahas-subramanian-frequenz (approved):

src/frequenz/channels/util/_select.py
```diff
@@ -106,6 +109,13 @@ def __del__(self) -> None:
         for task in self._pending:
             task.cancel()
 
+    async def stop(self) -> None:
+        """Stop the `Select` instance and cleanup any pending tasks."""
+        for task in self._pending:
+            task.cancel()
+        await asyncio.gather(*self._pending, return_exceptions=True)
```
leandro-lucarella-frequenz, line 116: Shouldn't this throw exceptions instead of returning them? Or this is just because you assume the only exception that could be captured is the `CancelledError` and this is why you are ignoring them? If so it might be worth a clarifying comment

Same for the other PR adding a stop().

Issue #73, opened 2023-02-17
Channel that doesn't remove the oldest messages when receiver buffer is full


ela-kotulska-frequenz:
### What's needed?

Now when receiver buffer size is full and new message comes, then the oldest message is removed.
But in some cases it is important that each request is received and processed.
One such use-case is when we request to subscribe for metric data.
If we have many components/metrics and we subscribe for all data at the startup, then default receiver buffer size is not enough. The oldest messages are removed and the application doesn't work properly because we don't receive data.

### Proposed solution

There are two possible solutions:
1. We could have special kind of blocking channel. 
If sender tries to send message, but the receiver buffer size is full, then we should block sender until the receiver buffer size has free space. For implementation simplicity it is enough if channel have many senders and only one receiver. (this covers our all current use-cases)

2. We could have option to dynamically increase the buffer size - but in that case there might be more and more requests, and receiver is unable to process them in time and the memory might increase too much. 

### Use cases

For now the only use-case was subscribing for the component metric data.

leandro-lucarella-frequenz:
Option 1 is #20, right?

Option 2 IMHO doesn't make a lot of sense, if you need a bigger buffer, just use a bigger buffer. If you mean an infinite buffer, I also don't think is a good idea.

llucax:
I will close this as a duplicate of #20 as I don't think option 2 is realistic.

PR #75, opened 2023-03-02
Experiment: A syncronous `send()`

mathias-baumann-frequenz:
This little experimental PR makes the `Sender.send()` method syncronous by using
`Event` instead of `Condition` as I saw no need to have an extra lock at that
place.

### DIFF
--- a/src/frequenz/channels/_base_classes.py
+++ b/src/frequenz/channels/_base_classes.py
@@ -18,7 +18,7 @@ class Sender(ABC, Generic[T]):
     """A channel Sender."""
 
     @abstractmethod
-    async def send(self, msg: T) -> None:
+    def send(self, msg: T) -> None:
         """Send a message to the channel.
 
         Args:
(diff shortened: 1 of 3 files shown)


ela-kotulska-frequenz:
It is very nice, but it is not enough.
We need blocking Sender anyway: https://github.com/frequenz-floss/frequenz-channels-python/issues/73
So this PR should be extended to make two interfaces, for blocking and non-blocking channels.

mathias-baumann-frequenz:
Adding a send that blocks on queue full shouldn't be a problem

mathias-baumann-frequenz:
But are you sure this should be in this PR? Combining a new feature with changing an existing one?

leandro-lucarella-frequenz:
I created an issue for this: #76

leandro-lucarella-frequenz:
> But are you sure this should be in this PR? Combining a new feature with changing an existing one?

Let's not change the `Sender` now, let's do the proper split in a separate PR, because probably all other channels senders should be sync too.

mathias-baumann-frequenz:
Closing this for now, the branch is still around.

PR #79, opened 2023-03-15
Make the timer generic and support periodic timers

### LINKED ISSUE #78: Add a periodic timer
### What's needed?

The current timer fires `interval` time after the `ready()` method was called, but this is not suitable for periodic timers as it will accumulate timer drift over time.

We need a timer that will periodically fire every `interval` time.

### Proposed solution

Extend `Timer` to have a `periodic` option or create a new `PeriodicTimer` that will start when the timer is instantiated and will fire periodically every `interval` time, no matter if `ready()` was called or not.

If the timer fired several times before `ready()`/`consume()` was called, those fires should be queued and returned eventually.

### Use cases

The resampler needs this.

One could use the current `Timer` and account for the time drifting oneself. But for that we need to create a new timer for each period, which is wasteful in terms of resources (allocations) and inconvenient in terms of programming effort.

### Additional context

* https://github.com/frequenz-floss/frequenz-sdk-python/issues/170

leandro-lucarella-frequenz:
The current timer provided is very convenient to implement timers, but it's not very helpful to implement a peridic timer that doesn't accumulate drift.

This PR makes the timer configurable via *missing ticks policies* and adds 3 policies: `TriggerAllMissing`, `SkipMissingAndResync` and `SkipMissingAndDrift`.

For convenience, 2 alternative constructors are provided for the most common cases: `Timer.timeout()` and `Timer.periodic()`. Timeout timers always drift and periodic timers never.

The new `Timer` also uses the `asyncio` loop monotonic clock, and returns the `drift` (difference between when the timer should have triggered and it actually did) instead of a `datetime` with the time of trigger.

It is also now possible to control when this timer is started, and can be stopped and reset/restarted at will.

Fixes #78.


leandro-lucarella-frequenz:
I'm marking this as a draft only because it is lacking tests, but it should be ready to review (I did some basic testing manually). I would like to get feedback before spending too much time with tests (as I would like to introduce also `async-solipsism` here).

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
...
+class PeriodicTimer(Receiver[timedelta]):
+    """A periodic timer receiver that fires every `interval` time.
+
+    The message it produces is a `timedelta` containing the drift of the timer,
+    i.e. the difference between when the timer should have fired and the time
+    when it actually fired. Note that the drift will not be accumulated by the
+    timer, which will be automatically adjusted so it only fires (or tries to)
+    in multiples of the `interval`.
+
+    This will not always be `0`, as if there is a task that is running when it
```
sahas-subramanian-frequenz:
```suggestion
    This will not always be `0`, because if there is a task that is running when it
```
because "as if" has another comparitive meaning, it is a bit confusing.

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
...
+
+        # If a stop was explicitly requested, we bail out.
+        if self._stopped:
+            return False
+
+        # If there is no stop requested but we don't have a time for the next
+        # message, then we reset() (so we start the timer as of now).
+        if self._next_msg_time is None:
+            self.reset()
+        assert self._next_msg_time is not None, "This should be assigned by reset()"
```
sahas-subramanian-frequenz: this assert can also be moved into the if block.

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
...
+
+    def consume(self) -> timedelta:
+        """Return the latest drift once `ready()` is complete.
+
+        Once the timer has fired (`ready()` is done), this method returns the
+        difference between when the timer should have fired and the time when
+        it actually fired. See the class documentation for more details.
+
+        Returns:
+            The time the timer drifted when waiting for the current interval.
```
sahas-subramanian-frequenz:
```suggestion
            The time by which the timer has drifted while waiting for the next tick.
```

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
...
+            self.reset()
+        assert self._next_msg_time is not None, "This should be assigned by reset()"
+
+        now = asyncio.get_running_loop().time()
+        diff = self._next_msg_time - now
+        if diff > 0:
+            await asyncio.sleep(diff)
+            now = asyncio.get_running_loop().time()
+        self._current_drift = timedelta(seconds=now - self._next_msg_time)
+        self._next_msg_time = self._next_msg_time + self._interval
```
sahas-subramanian-frequenz: if drift is greater than the interval,  do we want to directly skip over to the next tick that comes after the current time or do we make a tick for each missed interval?  I feel that we should just skip over the missed ones because we're just reporting clock drifts anyway, and not timestamps.

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
...
+
+        return True
+
+    def consume(self) -> timedelta:
```
sahas-subramanian-frequenz: What do you think about returning a `tuple[datetime, timedelta]`,  so people can get the next timestamp,  because most often that's what they need,  else they'll have to calculate it.  Or maybe just two timestamps,  and we can just have one timer `utils.Timer` instead:

``` python
class TimerTick:
    timer_period_ts: datetime
    timer_tick_ts: datetime
```

sahas-subramanian-frequenz (commented):
This looks good to me.  If we don't go the consolidating `TimerTick` route I mentioned,  we might want to add a docstring line here explaining the difference from `utils.Timer`,  and a line in `utils.Timer` saying when people should use a `PeriodicTimer` instead.

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: Maybe, I guess it was added to help `mypy`, I will try it.

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: Good question. I am emitting the tick because this was intended to be used by the resampler, and we want to keep resampling for all the periods even if they are late. Maybe we can make it a config option.

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: I didn't get the second part here. About returning `datetime` vs `timedelta` (or both), I know I thought about it and went with `timedelta`, but now I don't remember why.

This is a disadvantage of letting a PR sit there for so long, now I forgot all the context. I will get back to it and probably I will have more comments on your comments :)

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: OK, it seems to work too. Moved.

src/frequenz/channels/util/_periodic_timer.py
sahas-subramanian-frequenz: I was just proposing that we just have a single `utils.Timer` that returns both a periodic time, with uniform increments, and the tick time (i.e) clock time of the tick.

instead of having a separate class for each.  it would be the same performance-wise anyway.

leandro-lucarella-frequenz: I guess I put the delta because for the resampler I didn't need the timestamp, as I when the next resample was supposed to happen, but I guess other users might not be tracking this is it is good information to provide. The drift is used in the resampler to produce warnings if it is too big.

I will change to return `(supposed_fire_time, drift)`, I guess this is the most useful info, when it was supposed to fire (the *logical fire time*) and how it did differ from the actual time it fired. But actually any combination works, as it is possible to derive one from the other. Since our only use case so far has a case for the `(supposed_fire_time, drift)`, I would rather go with that than for what is needed for other hypothetical cases.

I still don't understand what you mean about this:

> Or maybe just two timestamps, and we can just have one timer utils.Timer instead:
>
> ```py
> class TimerTick:
>     timer_period_ts: datetime
>     timer_tick_ts: datetime
> ```

sahas-subramanian-frequenz: Maybe the field names weren't clear.  `utils.Timer` returns `actual_fire_time` only.  `utils.PeriodicTimer` returns `supposed_fire_time/drift`.  We can merge them both,  so there's no `PeriodicTimer`, only `Timer` that provides both types of info.

```python
class TimerTick:
    supposed_fire_time: datetime
    actual_fire_time: datetime
```

leandro-lucarella-frequenz: I see. I don't think it makes sense to merge them though, the way both work is very different, `Timer` will start *ticking* the moment you call `receive()` (or `ready()` technically), while the `PeriodicTimer` needs to be *ticking* all the time once it is started.

For a network timeout, `Timer` makes a lot of sense, so if we make it tick all the time as the periodic timer, we'll have to start/stop it manually to mimic the current behavior, which would make it harder to use.

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: I was thinking that since from now we only have a use case for the tick-always version, I would rather keep it like this for now and add the other option later, if needed.

src/frequenz/channels/util/_periodic_timer.py
sahas-subramanian-frequenz: ok, if we don't go the consolidation route, we might want to add a docstring line here explaining the difference from utils.Timer, and a line in utils.Timer saying when people should use a PeriodicTimer instead.

leandro-lucarella-frequenz: Now I remember why I was using only the drift, this is because we are using the monotonic time, so we have the time as a relative `float`, there is no timezone involved here, just an incrementing number. We need to create a datetime if we want to return one. I think we can do it, but since for my use case I didn't need it, I just preferred to go around this, as the user actually has all the information it needs. The user knows at what time the timer started, so it knows when the next tick is supposed to be due.

Also with the leap second it gets tricky, there is the theoretical scenario where as you might have a drift of 0.5, a supposed_fire_time of 12:00:00.00 and an actual_fire_time of 11:59:59.89 (but the drift was positive, it can never fire early). But we said this is rare enough (and it is supposed to go away) so I guess we can ignore it.

I will create the `datetime` anyway, but this adds an extra syscall to get the `datetime.now()` that didn't exit so far.

leandro-lucarella-frequenz: After looking at the code again I'm not that sure it is a good idea to return any `datetime`. The user already have all the info they need to track this themselves **iff** they need it, and I don't think many users of this class will actually need it, if you just want to run something periodically I don't see why we should assume the user cares at all about at what day or time this is triggering. If it were free to return this, I agree we should do it, this class doesn't know anything about `datetime` so far...

leandro-lucarella-frequenz: (again, looking at our only current use case, we don't need this. The resampler alrady knows what should be the next window timestamp and will always use that as a timestamp, no matter how much the timer has drifted, and it will only use the drift to show a warning if it is too big, so no need to call `datetime.now()` anywhere for this use case)

sahas-subramanian-frequenz: Do you need to call `datetime.now()` for every tick, if you just want to send the periodic time corresponding to the tick?

leandro-lucarella-frequenz:
Updated according to what we talked about. So now `PeriodicTimer` can have different behaviors when ticks are missed and can replace `Timer`, which I misunderstood how it worked. It is still a WIP but the implementation of the timer should be "finished" (I only need to review the docs). On some local tests it seems to work as expected.

I didn't include the `datetime` as we don't have any use cases for it yet and it is complex enough as it is, we can add it in the future if needed.

The tests are failing because of styling issues, it would be good if you can review it @sahas-subramanian-frequenz to see if you are good with it now so I can finish the docs, PR and removing `Timer`.

leandro-lucarella-frequenz:
Fixed some bugs and added some tests. I need to add some more but I consider this almost done. Not ready for merging because of the above, but not a draft anymore either.

leandro-lucarella-frequenz:
Since nobody looked at this yet, I'm converting to draft again because when doing more tests and using `hypothesis` to test better some cases, it become evident that using `float`s internally and doing modulo math and comparisons is very problematic in cases where values differ a lot in magnitude or significant digits, so I'm considering to use `int` internal representation (in `ms` or so).

I also don't like that `delay_tolerance` doesn't apply to every missed tick behavior, and that we need to handle every behavior in `ready()`, which is not scalable nor customizable, so I'll use classes to represent behaviors instead.

So a couple of major changes are coming.

leandro-lucarella-frequenz:
This PR is now ready for review for real. There are still quite a few commits that could be squashed together, but since this PR took so much work to get *right* (still to be seen :D), I'd prefer to keep the history as is, as it shows better the problems found while developing it.  I even wonder if the problems found here with `float`s might affect the SDK too, in particular classes doing alignment over `float`s (like the resampler).

Because of the above, I recommend to only go commit-by-commit to have a quick look to see if commits are consistent, but for the logic, it is probably better to review the final diff, because both the way the timer presents time internally and the missing tick policies were completely rewritten in the PR itself.

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,576 @@
...
+    """
+
+    def __init__(
+        self,
+        /,
+        interval: timedelta,
+        *,
+        auto_start: bool = True,
+        # We can use an instance here because TriggerAllMissed is immutable
+        missed_tick_policy: MissedTickPolicy = TriggerAllMissed(),
```
sahas-subramanian-frequenz: I think most of our actor use cases would use `SkipMissedAndResync`.  That should probably be the default behaviour.

sahas-subramanian-frequenz (dismissed):
Just one comment about default behaviour,  looks great otherwise. :-)

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: True. Even more, with the skip and drift policy the timer isn't really periodic anymore, I think what we have now is a generic timer. So I propose to rename the class to `Timer`, make the `missed_tick_policy` a required argument and add shortcut constructors for the most common use cases: `Timer.timeout(delay)` that returns a timer with skip and drift and 0 tolerance and `Timer.periodic(period, skip_missed_ticks=False)`, using trigger all or skip and resync policies as appropriate.

src/frequenz/channels/util/_periodic_timer.py
```diff
@@ -0,0 +1,201 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
daniel-zullo-frequenz: 2023

leandro-lucarella-frequenz:
Updated to implement this suggestion based on the comment from @sahas-subramanian-frequenz 

> True. Even more, with the skip and drift policy the timer isn't really periodic anymore, I think what we have now is a generic timer. So I propose to rename the class to `Timer`, make the `missed_tick_policy` a required argument and add shortcut constructors for the most common use cases: `Timer.timeout(delay)` that returns a timer with skip and drift and 0 tolerance and `Timer.periodic(period, skip_missed_ticks=False)`, using trigger all or skip and resync policies as appropriate.

Also added a commit to update the release notes to prepare it for the next release.

src/frequenz/channels/util/_timer.py
```diff
@@ -418,6 +419,97 @@ def __init__(
         if auto_start:
             self.reset()
 
+    @classmethod
+    def timeout(
+        cls,
+        delay: timedelta,
+        /,
```
daniel-zullo-frequenz, line 426: Should `/` go before `delay`?

src/frequenz/channels/util/_timer.py
```diff
@@ -418,6 +419,97 @@ def __init__(
...
+
+        This is basically a shortcut to create a timer with
+        `SkipMissedAndDrift(delay_tolerance=timedelta(0))` as the missed tick policy.
+
+        See the class documentation for details.
+
+        Args:
+            delay: The time until the timer ticks. Must be at least
+                1 microsecond.
+            auto_start: Wether the timer should be started when the
```
daniel-zullo-frequenz: s/wether/whether

src/frequenz/channels/util/_timer.py
```diff
@@ -418,6 +419,97 @@ def __init__(
...
+            SkipMissedAndDrift(delay_tolerance=timedelta(0)),
+            auto_start=auto_start,
+            loop=loop,
+        )
+
+    @classmethod
+    def periodic(
+        cls,
+        period: timedelta,
+        /,
```
daniel-zullo-frequenz, line 468: ditto if a previous similar comment applies

src/frequenz/channels/util/_timer.py
```diff
@@ -418,6 +419,97 @@ def __init__(
...
+        This is basically a shortcut to create a timer with either
+        `TriggerAllMissed()` or `SkipMissedAndResync()` as the missed tick policy
+        (depending on `skip_missed_ticks`).
+
+        See the class documentation for details.
+
+        Args:
+            period: The time between timer ticks. Must be at least
+                1 microsecond.
+            skip_missed_ticks: Wether to skip missed ticks or trigger them
```
daniel-zullo-frequenz: s/Wether/Whether/g (the typo is in 2 other places in the same file at least)

src/frequenz/channels/util/_timer.py
daniel-zullo-frequenz: this is actually correct so please discard this comment

src/frequenz/channels/util/_timer.py
daniel-zullo-frequenz: discarded

src/frequenz/channels/util/_periodic_timer.py
daniel-zullo-frequenz: This was fixed in another commit so feel free to discard it

src/frequenz/channels/util/_timer.py
```diff
@@ -1,135 +1,664 @@
...
+    Args:
+        time: The time to convert. If it is a `float`, it is assumed to be in
+            seconds.
+
+    Returns:
+        The time in microseconds.
+    """
+    if isinstance(time, timedelta):
+        time = time.total_seconds()
+    return round(time * 1_000_000)
```
daniel-zullo-frequenz, line 36: Just a louder thought about this function. The function is overwriting the input parameter `time` and it is doing a type conversion from `timedelta` to `float` if `time` is `timedelta`. The function is correct but generally speaking I don't think this is a good practice as it can lead to error/bugs

I'd suggest something around this:
```python
time_seconds: float
if isinstance(time, timedelta):
    time_seconds = time.total_seconds()
else:
    time_seconds = time
return round(time_seconds * 1_000_000)
```

or more pythonic

```python
time_seconds: float = time.total_seconds() if isinstance(time, timedelta) else time
return round(time_seconds * 1_000_000)
```

src/frequenz/channels/util/_timer.py
leandro-lucarella-frequenz: No, no, I definitely meant a castrated ram :goat: 

image

daniel-zullo-frequenz (commented):
I only have a cosmetic comment to check for and LGTM. Really good work with the documentation and examples which helps a lot  to the review process, also the changes well separated in different commits

src/frequenz/channels/util/_timer.py
leandro-lucarella-frequenz: No, I think you are looking at this from a more conventional static typed language perspective :)

Actually Python use "algebraic types" (I think that's how it is called). The type of `time` is `float | timedelta`, is both at the same time. But once you asserted somehow that the time is actually pinned to one of them, the type checker knows it and can still figure out the correct type. So I'm actually *narrowing* the type of `time` from `float | datetime` to `float` by converting the `datetime` to `float` when `time` is a `datetime`. Once the type is *narrowed*, the type system knows that `time` from now on is always `float`. This is why is not complaining when I'm calling `round()` (which takes a `float`) with `time`.

It would be something along these lines:

```python
    if isinstance(time, timedelta):
        time = time.total_seconds()
    assert isinstance(time, float)  # This is inferred by the type system
    time_float: float = time
    return round(time_float * 1_000_000)
```

So to say.  Does it make sense?

leandro-lucarella-frequenz:
Updated to remove all mentions to castrated ram (wether) :trollface:

src/frequenz/channels/util/_periodic_timer.py
leandro-lucarella-frequenz: Yeah, the first commits were a bit messy because it wasn't unclear how to solve the problem. It was too much work to split those commit properly so things ended up not super clean, for such a small thing I rather leave it as it is.

src/frequenz/channels/util/_timer.py
daniel-zullo-frequenz: > No, I think you are looking at this from a more conventional static typed language :)

Of course :)

> So to say. Does it make sense?

It does make sense and I just need to get used to

daniel-zullo-frequenz (approved):

sahas-subramanian-frequenz (approved):
I like the new naming scheme so much better than `PeriodicTimer` :tada:

leandro-lucarella-frequenz:
I know, and me too. Thanks for pushing back!

PR #83, opened 2023-04-14
Add support/tests for None-value channels

sahas-subramanian-frequenz:


src/frequenz/channels/_anycast.py
```diff
@@ -158,14 +159,18 @@ class Receiver(BaseReceiver[T]):
     method.
     """
 
+    @dataclass
+    class _Value(Generic[U]):
+        value: U
```
leandro-lucarella-frequenz: Maybe add a comment about why this wrapper is needed? I guess it is because `_next` being `None` has a special meaning, right? Another option would be to have a flag `_has_next`, but I'm good with this if the reason is more obvious.

sahas-subramanian-frequenz: This could be done with a sentinel,  but I was too lazy to look up the syntax.  Let me do that now.

sahas-subramanian-frequenz: I've added a fixup commit that switches to use of a sentinel.

src/frequenz/channels/_anycast.py
```diff
@@ -151,6 +151,16 @@ async def send(self, msg: T) -> None:
...
 
 
+class _Sentinel:
+    """A sentinel value to indicate that a value has not been set."""
+
+    def __repr__(self) -> str:
+        return "_Sentinel"
+
+    def __bool__(self) -> bool:
+        return False
```
leandro-lucarella-frequenz: Nitpicking here, but why are `__repr__` and `__bool__` needed here? Also I would call it `_Empty`, `_Absent` or something like that. It makes more obvious what is it for and `assert self._next is not _Empty` reads more easily than `assert self._next is not _Sentinel`.

leandro-lucarella-frequenz (dismissed):
Just a few final minor suggestions.

src/frequenz/channels/_anycast.py
sahas-subramanian-frequenz: `__repr__` is so that if someone tries to print it,  it doesn't say `<class something something>`,  but maybe not necessary here because it is not a general purpose empty sentinel.

`__bool__` is so we can treat it like None:

``` python
vv = _Empty()
if not vv:
   print("vv is empty")
```

I've renamed it to `_Empty`

leandro-lucarella-frequenz: But the problem with `bool` is that you first should make sure it is an `_Empty` instance, otherwise it could be a user sending a `False`, `None`, or anything that also supports `__bool__` as a valid message, so I'm not convinced is a good idea to even support it.

leandro-lucarella-frequenz (dismissed):
The only pending discussion is super minor so approving anyway in case you want to merge as is.

src/frequenz/channels/_anycast.py
sahas-subramanian-frequenz: `if not vv:` allows users to check for all False cases together,  for example for `list[int] | None | _Empty`.  Else they'll have to do: `if vv is not _Empty and vv is not None and len(vv) == 0:`

If they just want to check if it is empty,  they'd do `if vv is _Empty`,  just like `if vv is None`, etc if they just want to do those checks.

sahas-subramanian-frequenz: This is the Python way,  i.e like so many other Python idioms,  this is also unsafe and error-prone, and allows users to shoot themselves in the foot.

Okay,  I guess you're right.  I'll remove it then. :-)

sahas-subramanian-frequenz: done.

leandro-lucarella-frequenz (approved):

Issue #53, opened 2022-11-21
Make `FileWatcher` return the type of change observed


leandro-lucarella-frequenz:
Right now the `FileWatcher` only returns the path of the changed file, but it doesn't say if a change was a creation, modification or deletion, which might be useful information to have.

_Originally posted by @shsms in https://github.com/frequenz-floss/frequenz-channels-python/issues/51#issuecomment-1320171672_

PR #86, opened 2023-05-05
Fix `FileWatcher` `DELETE` event reporting and rearrange tests

leandro-lucarella-frequenz:
This pull request have 2 different but related main changes:

* The `FileWatcher` was skipping `DELETE` events, except on very weird cases where the file was deleted and re-created just before the event was triggered by `awatch()`.

  This also extends the integration test to test both situations (delete events when a file exist and when it doesn't).

  Because there is usually a considerable delay between a change in the filesystem happens and `awatch()` emits an event it is difficult to decide if we should filter events based on the file existing or not. For example, if a create event is received but before is triggered the file was removed, should we emit the create event or not? Or the other way around, should we emit a delete event if a file was created again before the event was triggered.

  It is probably best to leave this decision to the user, and let them deal with races.

* `FileWatcher` tests were rearranged and split between unit and integration tests.

  The existing tests were really integration tests, as they are not only testing `FileWatcher`, but also `Select` and `PeriodicTimer`, which is confusing when there is a bug in any of the other classes as we might get test failures here (this actually happened during the development of #79).
    
  Because of this the current tests were moved to `test/util/test_integration.py`. We also mark the tests with the `pytest` mark `integration` to make it more clear they are integration tests, and to enable an easy way to only run (exclude) integration tests.

  New unit tests were added to the current test file, properly mocking `awatch` to avoid interactions with the OS.

  There are many more integration tests posing as unit tests, but this PR only takes care of `FileWatcher`. With time we should move the rest too.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -2,7 +2,7 @@
 
 ## Summary
 
-This release adds support to pass `None` values via channels and revamps the `Timer` class to support custom policies for handling missed ticks and use the loop monotonic clock.
+This release adds support to pass `None` values via channels and revamps the `Timer` class to support custom policies for handling missed ticks and use the loop monotonic clock.  There is also a fix for the `FileWatcher` which includes a change in behavior when reporting changes for deleted files.
 
 ## Upgrading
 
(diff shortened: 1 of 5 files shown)


leandro-lucarella-frequenz:
Rebased and added a note on how to test the new timer with `async-solipsism` now.

I would make a release after this is merged.

RELEASE_NOTES.md
```diff
@@ -27,10 +27,32 @@ This release adds support to pass `None` values via channels and revamps the `Ti
 
   They are not **exactly** the same because the `triggered_datetime` in the second case will not be exactly when the timer had triggered, but that shouldn't be relevant, the important part is when your code can actually react to the timer trigger and to know how much drift there was to be able to take corrective actions.
 
-  Also the new `Timer` uses the `asyncio` loop monotonic clock and the old one used the wall clock (`datetime.now()`) to track time. This means that when using `async-solipsism` to test, the new `Timer` will always trigger immediately regarless of the state of the wall clock.
+  Also the new `Timer` uses the `asyncio` loop monotonic clock and the old one used the wall clock (`datetime.now()`) to track time. This means that when using `async-solipsism` to test, the new `Timer` will always trigger immediately regarless of the state of the wall clock.  This also means that we don't need to mock the wall clock with `time-machine` neither now.
```
daniel-zullo-frequenz: I think it should be `either` instead of `neither`, otherwise the sentence is double negative `we don't need...neither`

```suggestion
  Also the new `Timer` uses the `asyncio` loop monotonic clock and the old one used the wall clock (`datetime.now()`) to track time. This means that when using `async-solipsism` to test, the new `Timer` will always trigger immediately regarless of the state of the wall clock.  This also means that we don't need to mock the wall clock with `time-machine` either now.
```

RELEASE_NOTES.md
```diff
@@ -2,7 +2,7 @@
```
daniel-zullo-frequenz, line 2: The commit  `Fix FileWatcher skipping DELETE events` has comments that you might want to remove

daniel-zullo-frequenz (commented):
I have only 2 cosmetic comments to check for, LGTM otherwise

daniel-zullo-frequenz:
> I would make a release after this is merged.

@leandro-lucarella-frequenz can you please wait for a simple PR that addresses #53 which is based on the changes on this PR?

RELEASE_NOTES.md
leandro-lucarella-frequenz: Done

daniel-zullo-frequenz (approved):

PR #87, opened 2023-05-09
Report the type of change observed in FileWatcher

### LINKED ISSUE #53: Make `FileWatcher` return the type of change observed

daniel-zullo-frequenz:
Previously, the FileWatcher only provided information on the path of the changed file, without indicating whether the change was a creation, modification, or deletion. This commit improves the functionality of the FileWatcher by adding the ability to report the type of change observed. This additional information can be useful when tracking changes to files and understanding the context of the change.

Fixes #53

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -53,11 +53,18 @@ This release adds support to pass `None` values via channels and revamps the `Ti
 
   Therefore, you should now check a file receiving an event really exist before trying to operate on it.
 
+* `FileWatcher` reports the type of event observed in addition to the file path.
+
+  Previously, only the file path was reported. With this update, users can now determine if a file change is a creation, modification, or deletion.
+  Note that this change may require updates to your existing code that relies on `FileWatcher` as the new interface returns a `FileWatcher.Event` instead of just the file path.
+
 ## New Features
 
 * `util.Timer` was replaced by a more generic implementation that allows for customizable policies to handle missed ticks.
 
 * Passing `None` values via channels is now supported.
+
+* `FileWatcher.Event` was added to notify events when a file is created, modified, or deleted.
 
 ## Bug Fixes
 
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -8,7 +8,7 @@ This release adds support to pass `None` values via channels and revamps the `Ti
 
 * `util.Timer` was replaced by a more generic implementation that allows for customizable policies to handle missed ticks.
 
-  If you were using `Timer` to implement timeouts, these two pices of code should be almost equivalent:
+  If you were using `Timer` to implement timeouts, these two pieces of code should be almost equivalent:
 
   - Old:
 
@@ -27,7 +27,7 @@ This release adds support to pass `None` values via channels and revamps the `Ti
 
   They are not **exactly** the same because the `triggered_datetime` in the second case will not be exactly when the timer had triggered, but that shouldn't be relevant, the important part is when your code can actually react to the timer trigger and to know how much drift there was to be able to take corrective actions.
 
-  Also the new `Timer` uses the `asyncio` loop monotonic clock and the old one used the wall clock (`datetime.now()`) to track time. This means that when using `async-solipsism` to test, the new `Timer` will always trigger immediately regarless of the state of the wall clock.  This also means that we don't need to mock the wall clock with `time-machine` either now.
+  Also the new `Timer` uses the `asyncio` loop monotonic clock and the old one used the wall clock (`datetime.now()`) to track time. This means that when using `async-solipsism` to test, the new `Timer` will always trigger immediately regardless of the state of the wall clock.  This also means that we don't need to mock the wall clock with `time-machine` either now.
 
   With the previous timer one needed to create a separate task to run the timer, because otherwise it would block as it loops until the wall clock was at a specific time. Now the code will run like this:
 
@@ -39,7 +39,7 @@ This release adds support to pass `None` values via channels and revamps the `Ti
 
   # Simulates a delay in the timer trigger time
   asyncio.sleep(1.5)  # Advances the loop monotonic clock by 1.5 seconds immediately
-  await drift = timer.receive()  # The timer should have triggerd 0.5 seconds ago, so it doesn't even sleep
+  await drift = timer.receive()  # The timer should have triggered 0.5 seconds ago, so it doesn't even sleep
   assert drift == approx(timedelta(seconds=0.5))  # Now we should observe a drift of 0.5 seconds
   ```
 
--- a/src/frequenz/channels/util/_file_watcher.py
+++ b/src/frequenz/channels/util/_file_watcher.py
@@ -2,10 +2,12 @@
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
 """A Channel receiver for watching for new (or modified) files."""
+
+from __future__ import annotations
+
 import asyncio
 import pathlib
 from enum import Enum
-from typing import List, Optional, Set, Union
 
 from watchfiles import Change, awatch
 from watchfiles.main import FileChange
@@ -26,8 +28,8 @@ class FileWatcher(Receiver[pathlib.Path]):
 
     def __init__(
         self,
-        paths: List[Union[pathlib.Path, str]],
-        event_types: Optional[Set[EventType]] = None,
+        paths: list[pathlib.Path | str],
+        event_types: set[EventType] | None = None,
     ) -> None:
         """Create a `FileWatcher` instance.
 
@@ -48,8 +50,8 @@ class FileWatcher(Receiver[pathlib.Path]):
         self._awatch = awatch(
             *self._paths, stop_event=self._stop_event, watch_filter=self._filter_events
         )
-        self._awatch_stopped_exc: Optional[Exception] = None
-        self._changes: Set[FileChange] = set()
+        self._awatch_stopped_exc: Exception | None = None
+        self._changes: set[FileChange] = set()
 
     def _filter_events(
         self,
--- a/src/frequenz/channels/util/_file_watcher.py
+++ b/src/frequenz/channels/util/_file_watcher.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import asyncio
 import pathlib
+from dataclasses import dataclass
 from enum import Enum
 
 from watchfiles import Change, awatch
@@ -16,7 +17,7 @@ from .._base_classes import Receiver
 from .._exceptions import ReceiverStoppedError
 
 
-class FileWatcher(Receiver[pathlib.Path]):
+class FileWatcher(Receiver["FileWatcher.Event"]):
     """A channel receiver that watches for file events."""
 
     class EventType(Enum):
@@ -25,6 +26,15 @@ class FileWatcher(Receiver[pathlib.Path]):
         CREATE = Change.added
         MODIFY = Change.modified
         DELETE = Change.deleted
+
+    @dataclass(frozen=True)
+    class Event:
+        """A file change event."""
+
+        type: FileWatcher.EventType
+        """The type of change that was observed."""
+        path: pathlib.Path
+        """The path where the change was observed."""
 
     def __init__(
         self,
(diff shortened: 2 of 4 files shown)


src/frequenz/channels/util/_file_watcher.py
```diff
@@ -14,7 +14,7 @@
 from .._exceptions import ReceiverStoppedError
 
 
-class FileWatcher(Receiver[pathlib.Path]):
+class FileWatcher(Receiver[Tuple[Change, pathlib.Path]]):
```
leandro-lucarella-frequenz: This should actually return a `FileWatcher.EventType`, the underlying library used to implement `FileWatcher` is hidden to the user.

I would also recommended using a `dataclass` as the return:

```python
@dataclass(frozen=True)
class Event:
    type: EventType
    path: pathlib.Path
```

It avoids errors and makes the code more readable and future proof (if we add more attributes users don't need to update the code (if they are using tuple unpacking syntax, adding a new item to the tuple will make `(change, path) = fw.receive()` fail).

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -5,7 +5,7 @@
 import asyncio
 import pathlib
 from enum import Enum
-from typing import List, Optional, Set, Union
+from typing import List, Optional, Set, Tuple, Union
```
leandro-lucarella-frequenz: When you update old code to add more built-in types it would be nice if you convert it to using `from __future__ import annotations` instead. It is optional but very desirable so we can slowly get rid of the old idioms.

leandro-lucarella-frequenz (commented):
I propose some improvements to the API.

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: this suggestion makes totally sense, I'll update it accordingly. Thanks!

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: The Tuple will no longer be needed as we will use the `Event` class, however I'll add a commit to use annotations from __future__  instead

daniel-zullo-frequenz:
Updated based on a couple of suggestions from Luca.

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -14,7 +17,17 @@
 from .._exceptions import ReceiverStoppedError
 
 
-class FileWatcher(Receiver[pathlib.Path]):
+@dataclass(frozen=True)
+class FileEvent:
+    """A class to represent a file event useful for notifying file changes."""
+
+    type: Change
```
leandro-lucarella-frequenz: This still needs to be `FileWatcher.EventType`, not `Change`. We shouldn't expose anything from `watchfiles` to the user.

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -14,7 +17,17 @@
...
+    """The type of change observed."""
+    path: pathlib.Path
+    """The path where the change was observed."""
```
leandro-lucarella-frequenz: Any reason not to put `FileEvent` inside `FileWatcher`, just like `EventType`?

IIRC `EventType` was in the global scope too before, and I think it was changed as part of a restructuring to add the `util` package. I think I did this without thinking too much about google-style imports, and with that in mind it might not be the best approach. It is something we might need to review in the future. But for now I would try to keep the consistency.

Either both inside `FileWatcher` or both outside. About names, if it's inside, I think just plain `Event` is fine, because it must always be referred to by `FileWatcher.Event`. Even more, now the type could be nested in event: `FileWatcher.Event.Type`.

If it goes outside, I think `FileEvent` sounds good, then I would make `EventType` also global for consistency and call it `FileEventType`.

What I like about nesting is that is less import and the context of what each type is used for is super clear, but the downside is there might be more typing.

tests/utils/test_integration.py
```diff
@@ -10,8 +10,9 @@
 from datetime import timedelta
 
 import pytest
+from watchfiles import Change
```
leandro-lucarella-frequenz: This should never be imported in integration tests, as we shouldn't expose `watchfiles` to the end user.

daniel-zullo-frequenz: you're absolutely right. I should have used FileWatcher.EventType instead. Will update it

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: yeah my bad, will update it

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: > Any reason not to put FileEvent inside FileWatcher, just like EventType?

The reason was FileWatcher  is a `Receiver[FileEvent]` (inherits from). I would like to define Event internally if that's possible. I'll need to check it out

daniel-zullo-frequenz: Actually it works just quoting `FileWatcher.Event` in the class signature as follows:
```python
class FileWatcher(Receiver["FileWatcher.Event"]):
    ...
```

daniel-zullo-frequenz: Then I'll define `Event` inside `FileWatcher`

daniel-zullo-frequenz: done. Renamed `FileEvent` to `Event` and moved it inside `FileWatcher`

daniel-zullo-frequenz:
Updated based on latest suggestions from Luca

src/frequenz/channels/util/_file_watcher.py
leandro-lucarella-frequenz: Ok, interesting. It's surprising that it works as it is a class parametrized on itself 😱. I guess python can get away with a lot of crazy shit thanks to being so dynamic.

leandro-lucarella-frequenz (dismissed):

daniel-zullo-frequenz:
@leandro-lucarella-frequenz I added a trivial fix to correct typos and format in the releases notes (not related to this PR). Can you please check it and re-approve it if LGTY?

RELEASE_NOTES.md
```diff
@@ -8,16 +8,16 @@ This release adds support to pass `None` values via channels and revamps the `Ti
...
-  - Old:
+  * Old:
 
     ```python
     old_timer = Timer(1.0)
     triggered_datetime = old_timer.receive()
     ```
 
-  - New:
+  * New:
```
leandro-lucarella-frequenz: I started using different bullets types for different levels to make @shsms happy, I leave this to you, I don't care about either style :)

leandro-lucarella-frequenz (dismissed):
Just one super minor comment left, but approving anyway because it is just style and for a file that will be cleaned up right after this is merged.

RELEASE_NOTES.md
daniel-zullo-frequenz: hmm MD linter complains about it with: `MD004/ul-style: Unordered list style [Expected: asterisk; Actual: dash]markdownlint[MD004](https://github.com/DavidAnson/markdownlint/blob/v0.28.2/doc/md004.md)`

daniel-zullo-frequenz: ah I should be replaced by `+` instead to make Sahas happy :), if there is another subitem then it should be `-`
I'll change that

leandro-lucarella-frequenz: That's weird. It seems the default option is `consistent` and if I understand correctly this is consistent, or are there any other sub-lists using different bullets styles?

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -26,6 +27,15 @@ class EventType(Enum):
         MODIFY = Change.modified
         DELETE = Change.deleted
 
+    @dataclass(frozen=True)
+    class Event:
+        """A class to notify events when a file changes."""
```
Marenz: I don't think events need to be notified :)
Maybe just "Class used to notify about file changes"?

RELEASE_NOTES.md
daniel-zullo-frequenz: The linter still complains :(
I'll revert the format changes

leandro-lucarella-frequenz: Oh, no, it seems like with `sublist` option that would be valid. Not sure, as I said I don't care.  Whatever we decide it would be good to use that linter in the CI so we don't need to have these discussions anymore :D

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: Sounds good, will update it

RELEASE_NOTES.md
leandro-lucarella-frequenz: I would leave it as it is now to avoid the controversy in this PR. @daniel-zullo-frequenz  can you create a **Discussion** for this? Maybe in https://github.com/frequenz-floss/frequenz-repo-config-python/discussions/categories/ideas, so we just plug the linter in there directly.

src/frequenz/channels/util/_file_watcher.py
leandro-lucarella-frequenz: I don't get this, I don't read from there that events need to be modified. But now that you are moving my attention there, I think we should avoid starting class documentation with "A class to" and just say "A file change event." or something like that.

daniel-zullo-frequenz: done

RELEASE_NOTES.md
daniel-zullo-frequenz: o, i'll create a Discussion

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: I didn't see the last comment. @Marenz  does `A file change event.`  sound go to you?

RELEASE_NOTES.md
daniel-zullo-frequenz: Added here https://github.com/frequenz-floss/frequenz-repo-config-python/discussions/23

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -24,10 +27,19 @@ class EventType(Enum):
...
+        """The type of change that was observed."""
+        path: pathlib.Path
+        """The path where the change was observed."""
+
     def __init__(
         self,
-        paths: List[Union[pathlib.Path, str]],
-        event_types: Optional[Set[EventType]] = None,
+        paths: list[pathlib.Path | str],
+        event_types: set[EventType] | None = None,
```
Marenz, line 42: Might it be better to remove the `None` part and just default-init it with `set(FileWatcher.EventType)`? Then we can remove the `if` in line `51`

leandro-lucarella-frequenz: That's not a bad idea, but we would need to use `frozenset(FileWatcher.EventType)` then, because `set` is mutable and [the problem with mutable default arguments in Python](https://editorconfig.org/). If we use `frozenset()` for the default, then `event_types` type should be `collections.abc.Set` (or maybe `collections.abc.Iterable` would be enough if we convert it to a `frozenset` internally, then the argument type can be `Iterable` and the attribute type `frozenset`, so it can't be modified either via the attribute).

leandro-lucarella-frequenz (approved):
Besides agreeing to @Marenz suggestion, for me it is fine if it's left for a different PR so we can merge this sooner than later, as it is out of the scope of this PR.

src/frequenz/channels/util/_file_watcher.py
Marenz: Good observation!

daniel-zullo-frequenz:
> Besides agreeing to @Marenz suggestion, for me it is fine if it's left for a different PR so we can merge this sooner than later, as it is out of the scope of this PR.

Ok, I'll merge this PR as it currently is and create an issue to address https://github.com/frequenz-floss/frequenz-channels-python/pull/87#discussion_r1191102147

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: created the issue https://github.com/frequenz-floss/frequenz-channels-python/issues/88

PR #89, opened 2023-05-15
Update event_types parameter in FileWatcher

### LINKED ISSUE #88: Refactor event type and initialization in FileWatcher
### What's needed?

Change `event_type` type hint to remove `None`  and use the default-init `frozenset(FileWatcher.EventType)` instead.
Then the conditional check for `None` on `event_type` can be removed.

### Additional context

See review comment https://github.com/frequenz-floss/frequenz-channels-python/pull/87#discussion_r1191102147 for further details.

daniel-zullo-frequenz:
Previously, the event_types parameter in `FileWatcher` accepted either a `set` or `None` value, with `None` used as a mechanism to prevent issues related to Python mutable objects and default values.

This commit simplifies the code by removing the handling of `None` as a default value for event_types while still preventing issues related to mutable objects.

Fixes #88

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including if there are any depractions and what they should be replaced with -->
+* `FileWatcher` no longer accepts or sets `None` as the `event_types` argument. Instead, all available event types are now set by default while still providing the flexibility to customize the event types as needed.
 
 ## New Features
 
--- a/src/frequenz/channels/util/_file_watcher.py
+++ b/src/frequenz/channels/util/_file_watcher.py
@@ -7,6 +7,7 @@ from __future__ import annotations
 
 import asyncio
 import pathlib
+from collections import abc
 from dataclasses import dataclass
 from enum import Enum
 
@@ -39,19 +40,16 @@ class FileWatcher(Receiver["FileWatcher.Event"]):
     def __init__(
         self,
         paths: list[pathlib.Path | str],
-        event_types: set[EventType] | None = None,
+        event_types: abc.Iterable[EventType] = frozenset(EventType),
     ) -> None:
         """Create a `FileWatcher` instance.
 
         Args:
             paths: Paths to watch for changes.
-            event_types: Types of events to watch for or `None` to watch for
+            event_types: Types of events to watch for. Defaults to watch for
                 all event types.
         """
-        if event_types is None:
-            event_types = set(FileWatcher.EventType)  # all types
-
-        self.event_types = event_types
+        self.event_types: frozenset[FileWatcher.EventType] = frozenset(event_types)
         self._stop_event = asyncio.Event()
         self._paths = [
             path if isinstance(path, pathlib.Path) else pathlib.Path(path)
--- a/.github/RELEASE_NOTES.template.md
+++ b/.github/RELEASE_NOTES.template.md
@@ -6,7 +6,7 @@
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including if there are any depractions and what they should be replaced with -->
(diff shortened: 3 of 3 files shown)


RELEASE_NOTES.md
```diff
@@ -58,6 +58,8 @@ This release adds support to pass `None` values via channels and revamps the `Ti
   Previously, only the file path was reported. With this update, users can now determine if a file change is a creation, modification, or deletion.
   Note that this change may require updates to your existing code that relies on `FileWatcher` as the new interface returns a `FileWatcher.Event` instead of just the file path.
 
+* `FileWatcher` now accepts a`collections.abc.Iterable` object as `event_types` argument instead of a `set` or `None`, and defaults to a `frozenset` containing all available event types to ensure a consistent default behavior while enabling customizations.
```
leandro-lucarella-frequenz: Only the `None` change should be mentioned here, as it is the only change affecting the user. The change from `set` to `Iterable` is transparent for the user, or should be mentioned in the `New Features` as now more types can be used as an argument, but users don't need to change existing code.

Any reason not to take an `Iterator` instead of an `Iterable`? This would make it even more flexible (although in practice I doubt it makes any difference).

src/frequenz/channels/util/_file_watcher.py
```diff
@@ -39,19 +40,16 @@ class Event:
...
         paths: list[pathlib.Path | str],
-        event_types: set[EventType] | None = None,
+        event_types: abc.Iterable[EventType] = frozenset(EventType),
     ) -> None:
         """Create a `FileWatcher` instance.
 
         Args:
             paths: Paths to watch for changes.
-            event_types: Types of events to watch for or `None` to watch for
+            event_types: Types of events to watch for. Defaults to watch for
```
leandro-lucarella-frequenz, line 49: We don't really need to specify the default now, as the documentation renderer will show `frozenset(EventType)` as the default, but just FYI, I think in this case being more explicit is good as it is not trivial to see what `frozenset(EventType)` does.

leandro-lucarella-frequenz (commented):
Super minor comments, in general LGTM.

RELEASE_NOTES.md
daniel-zullo-frequenz: > Only the None change should be mentioned here, as it is the only change affecting the user. The change from set to Iterable is transparent for the user, or should be mentioned in the New Features as now more types can be used as an argument, but users don't need to change existing code.

Right, will update it

> Any reason not to take an Iterator instead of an Iterable? This would make it even more flexible (although in practice I doubt it makes any difference).

I've just tried it and `mypy` is not happy with it as it cannot assign the default  `frozenset` due to incompatibilities. See the error message below

```sh
nox > mypy --namespace-packages --non-interactive --install-types --explicit-package-bases --strict -p benchmarks -p docs -p frequenz.channels -p tests
src/frequenz/channels/util/_file_watcher.py:43: error: Incompatible default for argument "event_types" (default has type "FrozenSet[EventType]", argument has type "Iterator[EventType]")  [assignment]
src/frequenz/channels/util/_file_watcher.py:43: note: "frozenset" is missing following "Iterator" protocol member:
src/frequenz/channels/util/_file_watcher.py:43: note:     __next__
```

src/frequenz/channels/util/_file_watcher.py
daniel-zullo-frequenz: Yes, there was previously an inline comments to clarify that `set(EventType)` sets all event types so I thought it was worth it to explicitly mention it here. I'll keep it then if you don't have a strong opinion against it

RELEASE_NOTES.md
daniel-zullo-frequenz: Updated release notes

leandro-lucarella-frequenz: OK, I thought it was automatically converted to an iterator, but it seems it's not. :+1:

leandro-lucarella-frequenz (approved):

PR #92, opened 2023-05-16
Add dependabot config and pin dependencies

leandro-lucarella-frequenz:
This is a copy&paste from the SDK configuration:

- Add a dependabot config file
- Pin and narrow dependencies

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,24 @@
+version: 2
+updates:
+  - package-ecosystem: "pip"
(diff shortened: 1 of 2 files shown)


daniel-zullo-frequenz (approved):
The changes LGTM

Not related to this PR, IIUC it seems dependencies for testing or tools like mypy or pylint are not pinned in the way they are in the SDK which helps to have reproducible builds. If this assumption is right, we might need to address this issue in another PR

PR #114, opened 2023-06-02
Replace `Select` with `select()`, a type-safe async iterator

### LINKED ISSUE #27: Make `Select.ready` report one event at the time

### LINKED ISSUE #29: Improve `Select` documentation
### What's needed?

`Select` is supposed to be used in very specific cases, it is not really intended to be used everywhere as a general *event loop*, as when using `async` there is already an *event loop*.

Specially for people not used to use `async`, it might be tempting to use `Select` everywhere, leading to many issues, some of which can really be improved upon, but some that only come from the fact that `Select` is being misused.

Example issues that came from misusing `Select:

* https://github.com/frequenz-floss/frequenz-channels-python/issues/12
* https://github.com/frequenz-floss/frequenz-channels-python/issues/27
* https://github.com/frequenz-floss/frequenz-channels-python/discussions/24
* https://github.com/frequenz-floss/frequenz-sdk-python/pull/32#discussion_r1012610076

### Proposed solution

Improve the `Select` documentation to mention it should be only used in very niche situations.

#### Recommended usage

When receiving for multiple channels, normally it should be enough to create one task to handle each receiver, or use `Merge` or `MergeNamed` when multiple channels transport the same message type.

For example, we can find [this code in the SDK](https://github.com/frequenz-floss/frequenz-sdk-python/blob/1e01c2120cb96d703d3c3f6bb1fe1b738fa439d8/src/frequenz/sdk/data_ingestion/resampling/component_metrics_resampling_actor.py#L211-L240):

```py
            while await select.ready():
                if _ := select.resampling_timer:
                    awaitables = [
                        self._output_senders[channel_name].send(sample)
                        for channel_name, sample in self._resampler.resample()
                    ]
                    await asyncio.gather(*awaitables)
                if msg := select.component_data_receiver:
                    if msg.inner is None:
                        # When this happens, then DataSourcingActor has closed the channel
                        # for sending data for a specific `ComponentMetricRequest`,
                        # which may need to be handled properly here, e.g. unsubscribe
                        continue
                    channel_name, sample = msg.inner
                    if self.is_sample_valid(sample=sample):
                        self._resampler.add_sample(
                            time_series_id=channel_name,
                            sample=sample,
                        )
                if msg := select.subscription_receiver:
                    if msg.inner is None:
                        raise ConnectionError(
                            "Subscription channel connection has been closed!"
                        )
                    await self.subscribe(request=msg.inner)
                    # Breaking out from the loop is required to regenerate
                    # component_data_receivers to be able to fulfil this
                    # subscription (later can be optimized by checking if
                    # an output channel already existed in the `subscribe()` method)
                    break
```

Here `Select should not be used at all. There should be a task for the `resampling_timer` like:

```py
def resample_periodically(resampling_time):
    async for timestamp in resampling_timer:
        awaitables = [
            self._output_senders[channel_name].send(sample)
            for channel_name, sample in self._resampler.resample()
        ]
        await asyncio.gather(*awaitables)
```

Then similarly another task for `component_data_receiver`, or one task per resampler (and the resampler could even handle that task instead).

Finally the main loop should only handle the `subscription_receiver`, also using an async iterator is enough:

```py
    async for request in subscription_receiver:
        await self.subscribe(request)
```

Then we don't need a `break` to rebuild the `component_data_receiver` and the code becomes much more sequential an simpler to read and understand.

So as it can be seen here, misunderstanding `Select` could even lead to wrong design decisions.

`Select` should be used only in cases where there are **interactions** between different channels (and they are not all of the same type, so it is not enough to group them using `Merge` and `MergeNamed`. For example, we want to call `set_bounds()` each time a `bounds` message arrives, but if no messages arrive in 3 seconds, we want to call it too:

```py
timer = Timer(3.0)
select = Select(bounds=bounds_recv, timer=timer)
latest_bounds = {}

while await select.ready():
    if msg := select.bounds:
        latest_bounds[msg.inner.name] = msg.inner.value
        set_bounds(latest_bounds)
        timer.reset()
    elif select.timer:
        set_bounds(latest_bounds)
```

Here `bounds` and `timer` are interacting.

#### Design

We should also document that giving the user control over the order in which multiple ready receivers are processed is part of the design of the interface.

We also considered using an async iterator interface, but then the order in which multiple ready receivers are processed is in the hands of `asyncio.wait()` (or `as_completed()`, we tried both, and they result in different ordering patterns, as `as_completed()` seems to use first the first *awaitable* in the list, and `wait()` just returns a set with a *random* order).

This also can be bad for processing fairness, as we might end up with a very active receiver that always end up being the first ready receiver in the async iterator interface, leaving to starvation of all other receivers.

The alternative should be probably documented only in the code, but the user documentation should include and make it very clear that `Select` should be used when this ordering should be controlled by the user.

See #27 for more details.

### Use cases

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #47: Make `Select.ready()` an async iterable

### LINKED ISSUE #72: Find a way to make using select less likely to starve a receiver
Currently if `elif` or `break` is used when handling ready receivers is used, it could starve the receivers that come after it, which is not a good thing.

Ideally we should find a way to either forbid the use of `elif` or `break` or revisit #27.

_Originally posted by @leandro-lucarella-frequenz in https://github.com/frequenz-floss/frequenz-channels-python/pull/62#discussion_r1108415791_

leandro-lucarella-frequenz:
This new function acts as an async iterator, and makes sure no dangling tasks are left behind after a select loop is done.
    
This class is designed to replace the current `Select` implementation with the following improvements:
    
* Proper type hinting by using the new helper type guard `selected_from()`.
* Fixes potential starvation issues.
* Simplifies the interface by providing values one-by-one.
* Simplifies the implementation, so it is easier to maintain.
* Guarantees there are no dangling tasks left behind.

Here is an usage example:

```python
import datetime

from frequenz.channels import ReceiverStoppedError
from frequenz.channels.util import select, selected_from, Timer

timer1 = Timer.periodic(datetime.timedelta(seconds=1))
timer2 = Timer.timeout(datetime.timedelta(seconds=0.5))

async for selected in select(timer1, timer2):
    if selected_from(selected, timer1):
        if selected.was_stopped():
            print("timer1 was stopped")
            continue
        print(f"timer1: now={datetime.datetime.now()} drift={selected.value}")
        timer2.stop()
    elif selected_from(selected, timer2):
        if selected.was_stopped():
            print("timer2 was stopped")
            continue
        if exception := selected.exception is not None:
            print("timer2 was had an error")
            continue
        print(f"timer2: now={datetime.datetime.now()} drift={selected.value}")
    else:
        # This is not necessary, as select() will check for exhaustiveness, but
        # it is good practice to have it in case you forgot to handle a new
        # receiver added to `select()` at a later point in time.
        assert False
```


shsms:
I get the feeling that in cases like this, where we are replacing the entire implementation of a component, we could get creative about how we make the commits, such that it is easy to review.

For example,  
1. we could remove the old Select in the first commit,  add the new one in the second commit
2. add SelectNew as a separate implementation in the commit,  that will just replace `Select` in the second commit,  so we can spend much less time reviewing the second commit.

Just a suggestion,  and it also won't always be so straightforward.

src/frequenz/channels/util/_select.py
```diff
@@ -8,199 +8,223 @@
...
+                yield selected
+                assert selected._handled, f"{selected} was not handled"
+
+            for task in done:
+                receiver_active = task.result()
+                if not receiver_active:
+                    continue
+                name = task.get_name()
+                recv = receivers_map[name]
+                pending.add(_asyncio.create_task(_recv_task(recv), name=name))
```
shsms: why not a single loop?  going over the same set of items 3 times is slower than one right?  and that we can also eliminate the need for a `ready_set`.

src/frequenz/channels/util/_select.py
```diff
@@ -8,199 +8,223 @@
...
+        Returns:
+            A string representation of the instance.
+        """
+        return (
+            f"{self.__class__.__name__}({self._recv=}, {self._result=}, "
+            f"{self._exception=}, {self._handled=})"
         )
+
+
+def is_selected(
```
shsms: I guess this will change to `selected_from`?

shsms (commented):
Thanks Luca,  just looked at the essential parts of the `_select.py` file and all that looks good to me, except for a perf and a naming comment.

leandro-lucarella-frequenz:
> I get the feeling that in cases like this, where we are replacing the entire implementation of a component, we could get creative about how we make the commits, such that it is easy to review.
> 
> For example,
> 
>     1. we could remove the old Select in the first commit,  add the new one in the second commit
> 
>     2. add SelectNew as a separate implementation in the commit,  that will just replace `Select` in the second commit,  so we can spend much less time reviewing the second commit.
> 
> 
> Just a suggestion, and it also won't always be so straightforward.

Yeah, sorry if it was unclear, but the commits are not clean at all, I just put the draft out there to have a look mainly on how it is used and the new select implementation.

src/frequenz/channels/util/_select.py
leandro-lucarella-frequenz: It might be a leftover from a previous implementation.  At some point there was a reason for it, but maybe that reason doesn't exist any more.  I'll try to make all in one loop and if I can't I will comment on the why :+1:

llucax: Removed the extra loops :rocket:

llucax:
I updated the PR:

* Addressed @shsms comments (including better splitting of commits)
* The implementation itself now can be considered final, so it would be good to have a final review before I get into adding tests and replacing the current `Select` use in all the code.

TODO:
* Add tests
* Replace usage of the old `Select` with the new `select`
* Remove the old `Select`

llucax:
OK, here we go!  I wish this would not have taken so much time to get right (hopefully this time is right :crossed_fingers: ).

I decided to leave the failed `select()` function implementation for *historical reasons*. Also because at some point I doubted if it was a viable approach or not and even when now I'm pretty sure it is not, I wonder if Python will eventually change this. It is very weird, for one side it makes a lot of sense that a generator never regains control after a `async for` loop is broken, but for the other side it looks super broken that a `finally` statement can be skipped, so who knows...

llucax:
Oh, I need to add release notes :spiral_notepad:

llucax:
Added release notes.

src/frequenz/channels/util/_select.py
```diff
@@ -11,13 +11,598 @@
...
+# help of `assert_never` as described in:
+# https://docs.python.org/3.11/library/typing.html#typing.assert_never
+#
+# We also explored the possibility of using `match` to perform exhaustiveness checking,
+# but we couldn't find a way to make it work with `match`, and `match` is not yet
+# checked for exhaustiveness by `mypy` anyway, see:
+# https://github.com/python/mypy/issues/13597
+
+
+class Selector:
```
shsms: I think this should be renamed to `Select` because it is not used like a regular Python class which we create and call methods on.  It is used like a select function,  and is implemented as a class because of Python language limitations.  And it would then also be the same idiom from other languages.

shsms (dismissed):
Just a comment about the name,  which is of course subjective,  but looks good otherwise.

src/frequenz/channels/util/_select.py
llucax: I named it `Selector` because for me it is still mainly an async generator, not really a function, there is even no function call syntax involved to use it (except for creating the object). Even with the previous implementation it was weird to me to call it `Select`, because if was even more an object than this implementation, as you did have to call methods (like `ready()` on it).

Naming discussions are complicated and at the end I think the goal is to use whatever feels more obvious to users, so I will create a poll in our internal chat to see what people think is more readable/obvious, hope that is OK with you.

llucax: `select()` is back! The name discussion is moot now :tada:

llucax:
Updated to bring `select()` back! What a ride! :nauseated_face:

llucax:
Enabled auto-merge :scream:

shsms (approved):

PR #124, opened 2023-07-03
Migrate to use repo-config

llucax:
- Migrate to use repo-config
- Fix some typing issues in src/conftest.py
- Clear release notes


mkdocs.yml
```diff
@@ -102,6 +103,6 @@ plugins:
 
 # Preview controls
 watch:
-  - src
+  - "src"
```
Marenz, line 106: Why does src get quotes and the rest not?

Marenz:
In your [commit description](https://github.com/frequenz-floss/frequenz-channels-python/pull/124/commits/b8925c41bf5564188387dd47fec207a6b3b74879):

> These errors went **unnonticed** in the past because the file wasn't being
checked by mypy.

->

> These errors went **unnoticed** in the past because the file wasn't being
checked by mypy.

Marenz (dismissed):
I found a few minor things, I'll leave it up to you whether they should be ifxed

pyproject.toml
```diff
@@ -1,51 +1,95 @@
...
+  "darglint == 1.8.1",
+  "tomli == 2.0.1",      # Needed by pydocstyle to read pyproject.toml
+]
+dev-formatting = ["black == 23.3.0", "isort == 5.12.0"]
+dev-mkdocs = [
   "mike == 1.1.2",
   "mkdocs-gen-files == 0.5.0",
   "mkdocs-literate-nav == 0.6.0",
-  "mkdocs-material == 9.1.17",
+  "mkdocs-material == 9.1.16",
```
shsms: why going backwards?

mkdocs.yml
llucax: Because it is generated via a template and since we don't know the content 100% sure, we quote it. We don't need to have the inconsistency though, but it should be reported in repo-config, as the template should be fixed and then this repo updated via the template.

pyproject.toml
llucax: Oversight, the templates are outdated. I think it will be very annoying to keep templates up to date, so I thought it was better to just leave the projects to get updated via dependabot. But of course when migrating/updating a project we shouldn't revert dependencies, so I will fix, thanks!

llucax:
> I found a few minor things, I'll leave it up to you whether they should be ifxed

Thanks! Will fix the typo but I will leave the quotes to be fixed in repo-config.

llucax:
Updated!

llucax:
Enabled auto-merge.

Marenz (approved):

PR #136, opened 2023-07-13
Improve documentation

llucax:
- Add missing enum members documentation
- Make cross-linking more consistent
- Improve Timer documentation

### DIFF
--- a/src/frequenz/channels/util/_select.py
+++ b/src/frequenz/channels/util/_select.py
@@ -18,7 +18,7 @@ _T = TypeVar("_T")
 
 
 class Selected(Generic[_T]):
-    """A result of a [`select`][frequenz.channels.util.select] iteration.
+    """A result of a [`select()`][frequenz.channels.util.select] iteration.
 
     The selected receiver is consumed immediately and the received value is stored in
     the instance, unless there was an exception while receiving the value, in which case
@@ -28,7 +28,7 @@ class Selected(Generic[_T]):
     [`selected_from()`][frequenz.channels.util.selected_from] function to determine
     which receiver was selected.
 
-    Please see [`select`][frequenz.channels.util.select] for an example.
+    Please see [`select()`][frequenz.channels.util.select] for an example.
     """
 
     class _EmptyResult:
@@ -141,7 +141,7 @@ class Selected(Generic[_T]):
 def selected_from(
     selected: Selected[Any], receiver: Receiver[_T]
 ) -> TypeGuard[Selected[_T]]:
-    """Check if the given receiver was selected by [`select`][frequenz.channels.util.select].
+    """Check if the given receiver was selected by [`select()`][frequenz.channels.util.select].
 
     This function is used in conjunction with the
     [`Selected`][frequenz.channels.util.Selected] class to determine which receiver was
@@ -150,7 +150,7 @@ def selected_from(
     It also works as a [type guard][typing.TypeGuard] to narrow the type of the
     `Selected` instance to the type of the receiver.
 
-    Please see [`select`][frequenz.channels.util.select] for an example.
+    Please see [`select()`][frequenz.channels.util.select] for an example.
 
     Args:
         selected: The result of a `select()` iteration.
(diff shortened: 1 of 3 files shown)


src/frequenz/channels/util/_select.py
```diff
@@ -18,7 +18,7 @@
 
 
 class Selected(Generic[_T]):
-    """A result of a [`select`][frequenz.channels.util.select] iteration.
+    """A result of a [`select()`][frequenz.channels.util.select] iteration.
```
Marenz, line 21: So we always want to use `()` when referencing functions?

src/frequenz/channels/util/_timer.py
```diff
@@ -270,36 +270,49 @@ def __repr__(self) -> str:
 class Timer(Receiver[timedelta]):
     """A timer receiver that triggers every `interval` time.
 
-    The timer as microseconds resolution, so the `interval` must be at least
+    The timer as microseconds resolution, so the
```
Marenz: as -> has?

src/frequenz/channels/util/_timer.py
```diff
@@ -270,36 +270,49 @@ def __repr__(self) -> str:
...
-    when it actually triggered.
+    The message it produces is a [`timedelta`][datetime.timedelta] containing the drift
+    of the timer, i.e. the difference between when the timer should have triggered and
+    the time when it actually triggered.
 
     This drift will likely never be `0`, because if there is a task that is
     running when it should trigger, the timer will be delayed. In this case the
     drift will be positive. A negative drift should be technically impossible,
-    as the timer uses `asyncio`s loop monotonic clock.
+    as the timer uses [`asyncio`][asyncio]s loop monotonic clock.
```
Marenz, line 284: We seem to be using a mix of of `()` and `[]` here. Is `[name][ref]` different from `(name)[ref]`?

src/frequenz/channels/util/_timer.py
```diff
@@ -270,36 +270,49 @@ def __repr__(self) -> str:
...
 
-    If the timer is delayed too much, then the timer will behave according to
-    the `missed_tick_policy`. Missing ticks might or might not trigger
-    a message and the drift could be accumulated or not depending on the
-    chosen policy.
+    If the timer is delayed too much, then the timer will behave according to the
```
Marenz: > If the timer is delayed too much, then ~the timer~ **it** will

src/frequenz/channels/util/_select.py
llucax: In the rest of the file we always were referring to the select function as `select()`, so in this case I just wanted to have a consistent style.

src/frequenz/channels/util/_timer.py
llucax: `[name](ref)` is a regular, standard, markdown link, `[name][ref]` is a `mkdocstrings` extension to do code-crosslinking, so `ref` can only be a valid fully-qualified Python symbol.

llucax: When configuring [external "object inventories"](https://mkdocstrings.github.io/python/usage/#import), this references can even point to external python packages, in this case `asyncio` will link to Python's official docs.

src/frequenz/channels/util/_timer.py
llucax: Updated

src/frequenz/channels/util/_timer.py
llucax: Updated

src/frequenz/channels/util/_timer.py
llucax: BTW, `(name)[ref]` is incorrect, so if you spotted one is a bug, but I'm assuming is just a typo in your GitHub comment.

Marenz: This is one of my forever confusions in markdown syntax :sweat_smile: I can never remember the right order

Marenz (approved):

PR #154, opened 2023-08-17
Add Broadcast.wait_for_receiver() method

Marenz:


llucax:
What is the rationale for this?

Marenz:
I found it useful for tests especially, that you can start sending when you can be sure someone is reading, too.

llucax:
Any reason not to just first create the receivers and then the senders instead in a test? I really see no use case in production for this, the idea is that you don't care about this, a sender just sends, no matter if someone is listening or not. This kind of synchronization/dependency shouldn't be necessary by design, otherwise there is probably something wrong with the design... So I'm very hesitant to merge this.

llucax:
Also, you are not using anything private in the implementation, so there is nothing preventing users to do this in their code if it is really needed for testing.

Marenz:
> Any reason not to just first create the receivers and then the senders instead in a test? 

It's not always easily possible when your value travels through multiple receivers (component -> datasourcing actor -> resampling). 
Sometimes those receivers are only created just before you send a value (e.g. a component metric request is sent to the resampler, which sends it also to the data sourcing actor, once they are ready to stream the data, you can send values, but WHEN are they ready? On what should you wait after you created your own receiver on the exit end?

Issue #76, opened 2023-03-06
Implement `SyncSender` and `AsyncSender`


leandro-lucarella-frequenz:
### What's needed?

The `Sender` class provides an `async send()` method, ~but most (if not all) our senders could currently send synchronously~ [^1]. Being `async` means that the control flow/CPU could be taken away from the caller, which is very inconvenient for cases where sync is guaranteed.

[^1]: Actually, with the [removal of `Bidirectional`](https://github.com/frequenz-floss/frequenz-channels-python/issues/234), now we only have 2 types of channels and 1 is sync (broadcast) and the other one is async (anycast). There is also an [issue about having an option to make `Broadcast` senders block too](https://github.com/frequenz-floss/frequenz-channels-python/issues/20).

### Proposed solution

Split `Sender` into `SyncSender` and `AsyncSender` (probably make them [protocols](https://peps.python.org/pep-0544/) instead of classes). `SyncSender` would provide a sync `send()` and `AsyncSender` would provide a `async asend()`.

### Use cases

* Data sourcing actor: https://github.com/frequenz-floss/frequenz-sdk-python/pull/225#discussion_r1123163181
* The resampling actor (`Resampler` class) can also benefit from this, now the resampling functions need to be `async` because `send()` is `async`: https://github.com/frequenz-floss/frequenz-sdk-python/blob/ce6d19b869ed26c4df1e08a332f4e4fa4cbfc199/src/frequenz/sdk/timeseries/_resampling.py#L711-L716
* There could be channels that only provide a sync or an async sender interface.
* Broadcast channel could implement senders that support both operations, `send()` would drop like now and `asend()` would wait until the queue has room for the message.

### Alternatives and workarounds

Currently the Data sourcing actor is just spawning tasks for sending and assuming they will eventually finish, and successfully, as there is no error handling for the spawned tasks and if they don't end, we'll be leaking dangling tasks forever.

### Additional context

* Even if we don't have `AsyncSender`s now, we will probably need them in the future, at least for this issue: #20 
* Experiment using `sync send()` in the Data sourcing actor:
  * https://github.com/frequenz-floss/frequenz-sdk-python/pull/233
  * https://github.com/frequenz-floss/frequenz-channels-python/pull/75

mathias-baumann-frequenz:
Additional context: In the experimental branches, this change resulted in a speed up of the datasourcing benchmark from 30k/s to 70k/s samples

shsms:
[...]

PR #185, opened 2023-09-14
Document class and module attributes

### LINKED ISSUE #31: Add class/instance and module attributes/properties documentation
### What's needed?

We need to properly document class/instance and module attributes/properties documentation so they appear in the generated docs.

### Proposed solution

For now how to best do this is documented in this issue:

* https://github.com/frequenz-floss/frequenz-channels-python/pull/25

### Use cases

_No response_

_No response_

### Additional context

_No response_

daniel-zullo-frequenz:
Add or amend docstrings for all classes and modules where needed.

Fixes #31

### DIFF
--- a/src/frequenz/channels/_anycast.py
+++ b/src/frequenz/channels/_anycast.py
@@ -111,7 +111,7 @@ class Anycast(Generic[T]):
 class Sender(BaseSender[T]):
     """A sender to send messages to an Anycast channel.
 
-    Should not be created directly, but through the `Anycast.ggetet_sender()`
+    Should not be created directly, but through the `Anycast.new_sender()`
     method.
     """
 
@@ -211,7 +211,7 @@ class Receiver(BaseReceiver[T]):
 
         assert (
             self._next is not _Empty
-        ), "`consume()` must be preceeded by a call to `ready()`"
+        ), "`consume()` must be preceded by a call to `ready()`"
         # mypy doesn't understand that the assert above ensures that self._next is not
         # _Sentinel.  So we have to use a type ignore here.
         next_val: T = self._next  # type: ignore[assignment]
--- a/src/frequenz/channels/_anycast.py
+++ b/src/frequenz/channels/_anycast.py
@@ -68,10 +68,34 @@ class Anycast(Generic[T]):
             maxsize: Size of the channel's buffer.
         """
         self.limit: int = maxsize
+        """The maximum number of values that can be stored in the channel's buffer.
+
+        If the length of channel's buffer reaches the limit, then the sender
+        blocks at the [send()][frequenz.channels.Sender.send] method until
+        a value is consumed.
+        """
+
         self.deque: Deque[T] = deque(maxlen=maxsize)
+        """The channel's buffer."""
+
         self.send_cv: Condition = Condition()
+        """The condition to wait for free space in the channel's buffer.
+
+        If the channel's buffer is full, then the sender waits for values to
+        get consumed using this condition until there's some free space
+        available in the channel's buffer.
+        """
+
         self.recv_cv: Condition = Condition()
+        """The condition to wait for values in the channel's buffer.
+
+        If the channel's buffer is empty, then the receiver waits for values
+        using this condition until there's a value available in the channel's
+        buffer.
+        """
+
         self.closed: bool = False
+        """Whether the channel is closed."""
 
     async def close(self) -> None:
         """Close the channel.
@@ -79,7 +103,7 @@ class Anycast(Generic[T]):
         Any further attempts to [send()][frequenz.channels.Sender.send] data
         will return `False`.
 
-        Receivers will still be able to drain the pending items on the channel,
+        Receivers will still be able to drain the pending values on the channel,
         but after that, subsequent
         [receive()][frequenz.channels.Receiver.receive] calls will return `None`
         immediately.
@@ -122,6 +146,7 @@ class Sender(BaseSender[T]):
             chan: A reference to the channel that this sender belongs to.
         """
         self._chan = chan
+        """The channel that this sender belongs to."""
 
     async def send(self, msg: T) -> None:
         """Send a message across the channel.
@@ -169,6 +194,8 @@ class Receiver(BaseReceiver[T]):
             chan: A reference to the channel that this receiver belongs to.
         """
         self._chan = chan
+        """The channel that this receiver belongs to."""
+
         self._next: T | type[_Empty] = _Empty
 
     async def ready(self) -> bool:
--- a/src/frequenz/channels/_base_classes.py
+++ b/src/frequenz/channels/_base_classes.py
@@ -1,7 +1,7 @@
 # License: MIT
 # Copyright © 2022 Frequenz Energy-as-a-Service GmbH
 
-"""Baseclasses for Channel Sender and Receiver."""
+"""Base classes for Channel Sender and Receiver."""
 
 from __future__ import annotations
 
@@ -169,11 +169,13 @@ class _Map(Receiver[U], Generic[T, U]):
 
         Args:
             recv: The input receiver.
-            transform: The function to run on the input
-                data.
+            transform: The function to run on the input data.
         """
         self._recv = recv
+        """The input receiver."""
+
         self._transform = transform
+        """The function to run on the input data."""
 
     async def ready(self) -> bool:
         """Wait until the receiver is ready with a value or an error.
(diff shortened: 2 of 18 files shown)


src/frequenz/channels/_anycast.py
```diff
@@ -68,10 +68,19 @@ def __init__(self, maxsize: int = 10) -> None:
             maxsize: Size of the channel's buffer.
         """
         self.limit: int = maxsize
+        """The maximum number of items that can be stored in the channel's buffer."""
```
daniel-zullo-frequenz: ?
```suggestion
        """The maximum number of messages that can be stored in the channel's buffer."""
```

src/frequenz/channels/_anycast.py
```diff
@@ -68,10 +68,19 @@ def __init__(self, maxsize: int = 10) -> None:
...
+
         self.deque: Deque[T] = deque(maxlen=maxsize)
+        """The channel's buffer."""
+
         self.send_cv: Condition = Condition()
+        """The condition to wait for free space in the channel's buffer."""
+
         self.recv_cv: Condition = Condition()
+        """The condition to wait for data in the channel's buffer."""
```
daniel-zullo-frequenz: ?
```suggestion
        """The condition to wait for messages in the channel's buffer."""
```

src/frequenz/channels/_base_classes.py
```diff
@@ -169,11 +169,13 @@ def __init__(self, recv: Receiver[T], transform: Callable[[T], U]) -> None:
 
         Args:
             recv: The input receiver.
-            transform: The function to run on the input
-                data.
+            transform: The function to run on the input data.
         """
         self._recv = recv
+        """The input receiver."""
```
daniel-zullo-frequenz, line 175:
```suggestion
        """The input channel receiver."""
```

src/frequenz/channels/_anycast.py
llucax: We generally call it a `value` (what's sent through the channels).

I would also mention explicitly what happens if the buffer is full. Looking at the code it seems the sender will block at `send()` until a message is consumed. Maybe @shsms likes to confirm, but if this is the case, this is relevant to #76, as we already have some senders that can block (this one) and some that never do (broadcast).

src/frequenz/channels/_anycast.py
```diff
@@ -68,10 +68,19 @@ def __init__(self, maxsize: int = 10) -> None:
             maxsize: Size of the channel's buffer.
         """
         self.limit: int = maxsize
```
llucax, line 70: Damn, all of these should be private, or read-only at best. Not for this PR though:

* #186

src/frequenz/channels/_base_classes.py
llucax: I think just receiver is fine, as it could not be from a channel (like `Timer` or `FileWatcher`).

src/frequenz/channels/_bidirectional.py
```diff
@@ -117,21 +122,29 @@ def __init__(self, client_id: str, service_id: str) -> None:
             service_id: A name for the service end of the channels.
         """
         self._client_id = client_id
+        """The name for the client, used to name the channels."""
```
llucax, line 125: This is really confusing, but I know the original docs for `client_id` is and even the name of the variable.

But I guess we shouldn't fix the documentation, but fix the interface instead:

* #187

src/frequenz/channels/_broadcast.py
```diff
@@ -85,12 +85,22 @@ def __init__(self, name: str, resend_latest: bool = False) -> None:
                 to wait for the next message on the channel to arrive.
         """
         self.name: str = name
+        """The name of the broadcast channel."""
```
llucax: I would mention that this is for debugging purposes.

src/frequenz/channels/_broadcast.py
```diff
@@ -222,11 +233,19 @@ def __init__(self, uuid: UUID, name: str, maxsize: int, chan: Broadcast[T]) -> N
                 belongs to.
         """
         self._uuid = uuid
+        """The UUID to identify the receiver in the broadcast channel's list of receivers."""
+
         self._name = name
+        """The name to identify the receiver."""
```
llucax: I would also mention this is only for debugging.

src/frequenz/channels/_broadcast.py
```diff
@@ -222,11 +233,19 @@ def __init__(self, uuid: UUID, name: str, maxsize: int, chan: Broadcast[T]) -> N
...
+
         self._chan = chan
+        """The broadcast channel that this receiver belongs to."""
+
         self._q: Deque[T] = deque(maxlen=maxsize)
+        """The receiver's internal message queue."""
 
         self._active = True
+        """Whether the receiver is still active."""
```
llucax: Maybe mention what happens when it is inactive, or how/when it becomes inactive? After a quick look it seems the only way to make it inactive is by converting it into a peekable.

llucax (commented):
Just a few small comments and a few improvements that we might want to make before 1.0 as they are breaking changes :disappointed: (I created new issues for them).

src/frequenz/channels/_base_classes.py
daniel-zullo-frequenz: Good point, then will keep it as it is and resolving this conversation...

src/frequenz/channels/_anycast.py
daniel-zullo-frequenz: I see. Thanks for creating https://github.com/frequenz-floss/frequenz-channels-python/issues/186

src/frequenz/channels/_anycast.py
daniel-zullo-frequenz: btw `self.limit` is no longer used/needed so it can actually be deleted. Only `maxsize` is used to initialize the channel's buffer. I'll remove it if there is no objections.
And I'd extend the documentation for the blocking conditions instead

src/frequenz/channels/_broadcast.py
daniel-zullo-frequenz: done

src/frequenz/channels/_anycast.py
daniel-zullo-frequenz: Added a new commit to remove `self.limit`

src/frequenz/channels/_anycast.py
```diff
@@ -67,19 +67,35 @@ def __init__(self, maxsize: int = 10) -> None:
         Args:
             maxsize: Size of the channel's buffer.
         """
-        self.limit: int = maxsize
```
daniel-zullo-frequenz, line 70: this needs an entry into the release notes

RELEASE_NOTES.md
```diff
@@ -6,7 +6,7 @@ The `Timer` now can be started with a delay.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* The `limit` attribute was removed from `Anycast` channel as the same information can be obtained from its buffer through `deque.maxlen` property.
```
llucax: Why is this removed? If it is because of #186, then it should be made read-only, not removed, and I would leave it for a separate PR.

llucax (commented):
Not sure why did you remove the `limit` in the PR. The rest LGTM.

Also, could you not mark conversations I started as resolved? If you do so, I need to go expanding them one by one to see your replies and to remember what comments I made to go see in the code how it was addressed. If you need to keep track of which comments you already addressed, you can use a reaction instead, like :+1: .

RELEASE_NOTES.md
daniel-zullo-frequenz: It wasn't because of #186. Because of the comment I've found it redundant as the variable was only set and actually never used in the channel itself. Then exactly the same information can be retrieve from the `deque.maxlen` property. I've added a separate commit to remove it so I can easily discard the commit if you're against this change

llucax: I think in the future I wouldn't give access to users to the underlying `deque`, but it is nice to still provide a way to know what's the buffer size. In any case, I wouldn't decide this in this PR, we need to think about it a bit more before changing the API.

daniel-zullo-frequenz: fair enough, then I'll get rid of the related commits were `limit` was removed

daniel-zullo-frequenz: done

llucax (approved):
Thanks!

PR #189, opened 2023-09-18
docs: Improve README and intro

llucax:
- Add an intro and install instructions
- Improve the README introduction
- Add a quick start section to the README
- Add a documentation section to the README

### DIFF
--- a/docs/index.md
+++ b/docs/index.md
@@ -1,1 +1,80 @@
---8<-- "README.md"
+# Frequenz Channels for Python
+
+## Introduction
+
+Frequenz Channels is a *channels* implementation for Python.
+
+According to [Wikipedia](https://en.wikipedia.org/wiki/Channel_(programming)):
+
+> A channel is a model for interprocess communication and synchronization via
+> message passing. A message may be sent over a channel, and another process or
+> thread is able to receive messages sent over a channel it has a reference to,
+> as a stream. Different implementations of channels may be buffered or not,
+> and either synchronous or asynchronous.
+
+Frequenz Channels are mostly designed after [Go
+channels](https://tour.golang.org/concurrency/2) but it also borrows ideas from
+[Rust channels](https://doc.rust-lang.org/book/ch16-02-message-passing.html).
+
+## Installation
+
+First, you need to make sure you have Python installed (at least version 3.11):
+
+```console
+$ python3 --version
+Python 3.11.4
+```
+
+!!! note
+
+    These instructions assume you are using a [POSIX compatible
+    `sh`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html)
+    shell.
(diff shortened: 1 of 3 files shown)


docs/index.md
```diff
@@ -1 +1,80 @@
...
+```
+
+!!! note
+
+    These instructions assume you are using a [POSIX compatible
+    `sh`](https://pubs.opengroup.org/onlinepubs/9699919799/utilities/sh.html)
+    shell.
+
+If that command doesn't print a version newer than 3.11.0, you'll need to
+[download and install Python](https://www.python.org/downloads/) first.
```
daniel-zullo-frequenz, line 35: the instructions rely on having `pip` and `venv` already installed, and these modules might not be installed if the user  only install python3.11. So it might be worth mentioning the installation of `pip` and `venv` is also needed

daniel-zullo-frequenz (commented):
Only a comment to check for, LGTM otherwise

docs/index.md
llucax: You are right about `pip`, but `venv` is a [built-in Python module](https://docs.python.org/3/library/venv.html), are you sure it is not installed by default? That would be very weird...

llucax: Added link to instruction on how to install `pip`.

daniel-zullo-frequenz: I've tested quickly on docker using latest ubuntu and I've got the following message:

```
The virtual environment was not created successfully because ensurepip is not
available.  On Debian/Ubuntu systems, you need to install the python3-venv
package using the following command.

    apt install python3.11-venv
```

but that does not mean it is the same for all environments.
Then I think having instructions for only `pip` is enough

daniel-zullo-frequenz (approved):

docs/index.md
daniel-zullo-frequenz: extracted from https://packaging.python.org/en/latest/guides/installing-using-pip-and-virtual-environments/
 ```
Installing virtualenv
Note If you are using Python 3.3 or newer, the [venv](https://docs.python.org/3/library/venv.html#module-venv) module is the preferred way to create and manage virtual environments. venv is included in the Python standard library and requires no additional installation.
```

llucax: I remembered stumbling on that issue in the past too, but I think it is a problem with how multiple python versions are installed in Ubuntu., when installing from sources for sure it comes with venv. Also, if you are in Ubuntu, it looks like it tells you how to fix it. On the other hand, using venv is not technically necessary, it is just a recommendation.

daniel-zullo-frequenz: > Also, if you are in Ubuntu, it looks like it tells you how to fix it. On the other hand, using venv is not technically necessary, it is just a recommendation.

right, I have a similar thought about it.

PR #213, opened 2023-10-16
Make internal variable names private

### LINKED ISSUE #186: Many class attributes should be private or read-only
### What's needed?

We have a lot of class attributes that are public, so users can mess with them even when it is not safe to do so.

Just as an example, the `Anycast` channel has:

```python
        self.limit: int = maxsize  # If changed, the deque is not updated
        self.deque: Deque[T] = deque(maxlen=maxsize) # a user could just add or remove stuff without using the condition variable
        self.send_cv: Condition = Condition()  # users could triggers sends when the deque is full, dropping messages unexpectedly
        self.recv_cv: Condition = Condition()
        self.closed: bool = False # users could mark the channel as open after it was closed.
```

### Proposed solution

Review all class attributes and make them private or read-only as appropriate.

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
The `Anycast` and `Broadcast` implementations were exposing internal
variables to users.  This shouldn't be the case and is fixed in this
PR.

Closes #186

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -6,7 +6,7 @@ The `Timer` now can be started with a delay.
 
 ## Upgrading
 
-<!-- Here goes notes on how to upgrade from previous versions, including deprecations and what they should be replaced with -->
+* Internal variable names in the `Anycast` and `Broadcast` implementations are now private.
 
 ## New Features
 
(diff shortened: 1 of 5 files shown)


daniel-zullo-frequenz (commented):
LGTM. Can you please check if this PR addresses https://github.com/frequenz-floss/frequenz-channels-python/issues/186 and mention it in the PR description if that's the case?

shsms:
> Can you please check if this PR addresses https://github.com/frequenz-floss/frequenz-channels-python/issues/186 and mention it in the PR description if that's the case?

done

daniel-zullo-frequenz (approved):

PR #221, opened 2023-10-25
Make `resend_latest` a public attribute for `Broadcast` channels

shsms:
This allows modifying the flag after the channel is constructed, for
example in dynamically created channels.

### DIFF
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -75,14 +75,15 @@ class Broadcast(Generic[T]):
         """Create a Broadcast channel.
 
         Args:
-            name: A name for the broadcast channel, typically based on the type
-                of data sent through it.  Used to identify the channel in the
-                logs.
+            name: A name for the broadcast channel, typically based on the type of data
+                sent through it.  Used to identify the channel in the logs.
             resend_latest: When True, every time a new receiver is created with
-                `new_receiver`, it will automatically get sent the latest value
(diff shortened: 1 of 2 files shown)


src/frequenz/channels/_broadcast.py
```diff
@@ -90,7 +90,7 @@ def __init__(self, name: str, resend_latest: bool = False) -> None:
         Only used for debugging purposes.
         """
 
-        self._resend_latest = resend_latest
+        self.resend_latest = resend_latest
         """Whether to resend the latest value to new receivers."""
```
daniel-zullo-frequenz: Optionally, I'd add a similar note to the one added in the SDK

`resend_latest` is `False` by default.  It is safe to be set in data/reporting
channels, but is not recommended for use in channels that stream control
instructions.

daniel-zullo-frequenz (commented):
I have an optional suggestion to check for and LGTM

src/frequenz/channels/_broadcast.py
shsms: Yes,  I was thinking that it was not relevant in the channel library, because from this point of view, anything is just a message.

But maybe it would be helpful to mention, will do.

shsms: done

daniel-zullo-frequenz: Thanks!

daniel-zullo-frequenz (approved):

PR #230, opened 2023-11-01
Update to repo-config v0.7.2

llucax:
- Fix the wrong `cli` package name.
- Improve the arm64 `Dockerfile` description.

### DIFF
--- a/.github/workflows/ci.yaml
+++ b/.github/workflows/ci.yaml
@@ -363,7 +363,7 @@ jobs:
         if: steps.mike-version.outputs.version
         run: |
           git checkout gh-pages
-          python -m frequenz.repo.config.cli.version.mkdocs.sort versions.json
+          python -m frequenz.repo.config.cli.version.mike.sort versions.json
           git commit -a -m "Sort versions.json"
 
       - name: Publish site
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -5,7 +5,7 @@
 requires = [
   "setuptools == 68.1.0",
   "setuptools_scm[toml] == 7.1.0",
-  "frequenz-repo-config[lib] == 0.7.1",
+  "frequenz-repo-config[lib] == 0.7.2",
 ]
 build-backend = "setuptools.build_meta"
 
@@ -53,7 +53,7 @@ dev-mkdocs = [
   "mkdocs-material == 9.3.2",
   "mkdocs-macros-plugin == 1.0.4",
   "mkdocstrings[python] == 0.23.0",
-  "frequenz-repo-config[lib] == 0.7.1",
+  "frequenz-repo-config[lib] == 0.7.2",
 ]
 dev-mypy = [
   "mypy == 1.6.1",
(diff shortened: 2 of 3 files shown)


llucax:
Blocked by:

* https://github.com/frequenz-floss/frequenz-repo-config-python/pull/167
* `repo-config` v0.7.2 release

.github/workflows/ci.yaml
```diff
@@ -363,7 +363,7 @@ jobs:
         if: steps.mike-version.outputs.version
         run: |
           git checkout gh-pages
-          python -m frequenz.repo.config.cli.version.mkdocs.sort versions.json
+          python -m frequenz.repo.config.cli.version.mike.sort versions.json
```
christianparpart, line 366: I would not have spotted that! I mean, what's `mike`?

pyproject.toml
```diff
@@ -71,7 +71,7 @@ dev-pytest = [
   "pytest == 7.4.3",
   "async-solipsism == 0.5",
   "hypothesis == 6.88.1",
-  "frequenz-repo-config[extra-lint-examples] == 0.7.1",
+  "frequenz-repo-config[extra-lint-examples] == 0.7.2",
```
christianparpart, line 74: What generally bugs me (and that's not only a Python-only problem), is that you need to update the version in so many places, all at once, that it's not too hard to accidentally miss one, when you want to bump it. But I guess there's no better way of doing it, or is there?

What I imagine in my head is some kind of single source of truth, and if dev-pylint as well as dev-noxfile both need frequenz-repo-config at version X, then they $var that version from that single source. This would make upgrading at least way less error-prone. (I'm just thinking out loud here) :)

christianparpart (approved):

.github/workflows/ci.yaml
llucax: https://github.com/frequenz-floss/frequenz-repo-config-python/blob/v0.x.x/CONTRIBUTING.md#building-the-documentation 

or https://github.com/jimporter/mike

pyproject.toml
llucax: Not that I know of, we can greatly simplify this by having only one `dev` optional dependencies that just install everything, that's in the roadmap for `repo-config`, I think having this level of fine grain have very little advantages and a lot of issues. But we'll still need to put it twice (one in build deps and one in dev deps).

llucax: The other issue is repo-config now holds both a library and the cookiecutter templates. Actually all these point releases are for the cookiecutter templates only, so there is no difference if we didn't upgrade the dependency here. Once we split the templates to a different repo, this should also become better.

Then we might even want to split repo-config in the *build* part of it and the *dev* part of it, so then both dependencies would be independent and we'll only need to update one when it really changes (but I see this as a second stage change).

Finally, once it is more stable, dependabot should be able to update repo-config automatically too.

But yeah, a lot could be done to improve the situation but it requires a lot of work.

.github/workflows/ci.yaml
christianparpart: thx

PR #231, opened 2023-11-02
Clean up and improve the code and public interface

### LINKED ISSUE #187: Consolidate and make channels `name`s (`client_id` / `service_id`) optional
### What's needed?

A clean and consistent interface to create channels, and one that doesn't require to pass strings used for debugging only.

### Proposed solution

- [ ] Join both `client_id` and `service_id` from `Anycast` constructor into an `name` argument to match `Broadcast`.
- [ ] Make `name` optional

  If `name` is `None` I would just generate a [`uuid`](https://docs.python.org/3/library/uuid.html) or just `id()` (in a distributed context, `uuid` will make any unlikely collision even more unlikely).

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
This PR addresses many small cleanups and improvements to the code and the public interface, including some minor breaking changes.

Cleanups:

- Use `deque` for typing
- Replace `__anext__()` with `anext()`
- Rename `recv` argument as `receiver`
- Add missing types to members in `__init__()`
- Anycast: Don't store the `_limit`
- Use the string representation for logging
- Broadcast: Move the `resend_latest` attribute to the end
- Remove `logging` from `Timer` examples
- Make `Broadcast` logger private

Improvements:

- Add `is_closed` property to channels
- Anycast: Clarify what happens when the buffer is full
- Anycast: Add `name` argument
- Add `name` property to bidirectional
- Broadcast: Use `hash()` instead of a UUID
- Event: Make the default name more readable
- Add descriptive `str` and `repr` implementations
- Bump `repo-config` to v0.7.5
- Bump `pylint` to 3.0.2
- Bump `mike` to 2.0.0
- Anycast: Log a warning when a sender is blocked

Breaking changes:

- Make some arguments keyword-only
- Make exception messages `str`
- Rename `maxsize` to `limit`
- Broadcast: Make `name` keyword-only
- Bidirectional: unify `client_id` and `service_id` (fixes #187).


src/frequenz/channels/_broadcast.py
```diff
@@ -161,15 +159,15 @@ def new_receiver(
 
         Args:
             name: A name to identify the receiver in the logs.
-            maxsize: Size of the receiver's buffer.
+            limit: Size of the receiver's buffer in number of messages.
```
shsms: Maybe this is better?

```suggestion
            limit: Number of messages the receiver can hold in its buffer.

File: src/frequenz/channels/_broadcast.py [hunk 2]
```diff
@@ -71,12 +71,13 @@ async def recv(id: int, receiver: channel.Receiver) -> None:
         Check the `tests` and `benchmarks` directories for more examples.
     """
 
-    def __init__(self, name: str, *, resend_latest: bool = False) -> None:
+    def __init__(self, *, name: str | None, resend_latest: bool = False) -> None:
```
shsms: I'm not so sure about making the name parameter of `Broadcast` channels optional, because people would just create channels without thinking of names, and we'll have logs complaining about random channels filling up, and it becomes super hard to find out where the bottle neck is.

src/frequenz/channels/_broadcast.py
```diff
@@ -127,6 +127,15 @@ def name(self) -> str:
         """
         return self._name
 
+    @property
+    def latest(self) -> T | None:
```
shsms: Not sure it is a good idea to add this directly on the channel, unless you have a clear usecase in mind already?

shsms (commented):
Just a few comments.

Also looks like some commented lines got into the commit message here, I wonder how: https://github.com/frequenz-floss/frequenz-channels-python/pull/231/commits/25a9099f6ebd79046a215281bf6c3c70a5d6311d

src/frequenz/channels/_broadcast.py
christianparpart: Why do you prefer `A | None` over `Optional[A]`? I am asking purely out of curiosity. :)

llucax:
> Also looks like some commented lines got into the commit message here, I wonder how: [25a9099](https://github.com/frequenz-floss/frequenz-channels-python/commit/25a9099f6ebd79046a215281bf6c3c70a5d6311d)

Oh, good catch, thanks, it sometimes happen when using the scissors style for comments, I use it to avoid messing with markdown in Git messages.

src/frequenz/channels/_broadcast.py
llucax: Not really strong reasons, but no need for an extra `import` and `Optional` was just a library solution until `|` got properly implemented in Python 3.10.

llucax: > I'm not so sure about making the name parameter of `Broadcast` channels optional, because people would just create channels without thinking of names, and we'll have logs complaining about random channels filling up, and it becomes super hard to find out where the bottle neck is.

Mmm, I see. For me it was weird to require an argument that it is only used for logging, but I can see how otherwise people might need to learn the hard way to use custom names.

Also my feeling is, this is only a special case now in `Broadcast` because is what we use the most, but following that argument, `Anycast` should probably also log when a sender is being blocked because the buffer is full, as it is also a sign of receivers not being able to keep up.

Also, this might apply to other classes too, like `Merge`or whatever, where you could potentially have issues with one instance and that get logged and then you don't know which one is giving issues (although I know this is more a theoretical issue). But since this is an interface issue, and it is hard to know what future requirements would be, if names are really important we might think about making them required even if they are mostly unused for now (except for the `str`/`repr` implementation).

I know the above might be a bit extreme, but at least for channel classes, I think we should have a consistent approach and have all names optional or required.

src/frequenz/channels/_broadcast.py
llucax: No, not in particular, it just entered in the bag of private stuff that should be safe to expose as read-only. But thinking about the case we were talking today about only needing the last value, I was sort of wondering why even a receiver is needed at all if we can access the `latest` value from the channel directly.

I guess your answer will be something like "because the channel is sort of a private interface and should be used through senders and receivers", not sure what's the right balance between pragmatism and purism :D

src/frequenz/channels/_broadcast.py
shsms: Yes, I agree on consistency, and we need it to be mandatory only for the channels, because everything else can derive the names for logging from the channel.

src/frequenz/channels/_broadcast.py
llucax: Updated with the suggestion.

src/frequenz/channels/_broadcast.py
shsms: > I was sort of wondering why even a receiver is needed at all if we can access the latest value from the channel directly.

yes, that's exactly what a `Peekable` does.  And the `_latest` variable was added to support the `Peekable` implementation, so I'd rather remove that as well.

And not all channels are accessible.  We have the `new_receiver` interface everywhere in the sdk's data pipeline, and even in the microgrid client.  Those classes own the channels and give out only receivers.  So it is also easier to have the same interface everywhere.

llucax:
> Oh, good catch, thanks, it sometimes happen when using the scissors style for comments, I use it to avoid messing with markdown in Git messages.

Fixed in the update too.

src/frequenz/channels/_broadcast.py
shsms: And it is one less hurdle once we start trying to go to multi-process world.

src/frequenz/channels/_broadcast.py
llucax: OK, I will make it mandatory for all channels then.

src/frequenz/channels/_broadcast.py
llucax: OK, I will remove it.

llucax:
Updated to remove the `Broadcast.latest` property and to make channel names required.

Also added these 3 extra commits cleaning up logging a bit and adding a new log when `Anycast` senders are blocked as previously suggested:

* Remove `logging` from `Timer` examples
* `Anycast`: Log a warning when a sender is blocked
* Make `Broadcast` logger private

src/frequenz/channels/_broadcast.py
llucax: Done

src/frequenz/channels/_broadcast.py
llucax: Done

shsms (approved):

PR #238, opened 2023-11-09
Replace `Merge` and `MergeNamed` with `merge()`

### LINKED ISSUE #236: Remove `MergeNamed`
### What's needed?

We already have 2 ways to join multiple receivers, `Merge` and `select`. `MergeNamed` is a mixture of both, so it is redundant.

### Proposed solution

Remove `MergeNamed`.

### Use cases

_No response_

_No response_

### Additional context

_No response_

### LINKED ISSUE #237: Expose `Merge` as a `merge()` function
### What's needed?

There is not much use in having `Merge` as a class from a user point of view, users normally only care about merging some channels, so having a class just makes the syntax more weird for most use cases that will be `async for msg in merge(...)`.

### Proposed solution

Make `Merge` private (`_Merge`) and create a new function that just does `return _Merge(...)`.

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
- Remove `MergeNamed`
- Add a `merge()` wrapper function that replaces `Merge`
- Rename `Merge` to `_Merge` to make it private more explicitly
- Rename `_Merge` variable `args` to `receivers`
- Improve string representation of merge receiver
- Raise an exception is `merge()` is called with less than two receivers

Fixes #236, fixes #237.


src/frequenz/channels/__init__.py
```diff
@@ -26,11 +26,10 @@
 
 Utilities to work with channels:
 
-* [Merge][frequenz.channels.Merge] and [MergeNamed][frequenz.channels.MergeNamed]:
-  [Receivers][frequenz.channels.Receiver] that merge messages coming from multiple
-  receivers into a single stream.
+* [merge][frequenz.channels.merge]: Merges messages coming from multiple receivers into
+  a single stream.
```
daniel-zullo-frequenz: To be consistent with `select` documentation
```suggestion
* [merge][frequenz.channels.merge]: Merge messages coming from multiple receivers into
  a single stream.
```

src/frequenz/channels/_merge.py
```diff
@@ -13,13 +13,23 @@
...
+
+    Args:
+        *receivers: The receivers to merge.
+
+    Returns:
+        A receiver that merges the messages coming from multiple receivers into a
+            single stream.
+
+    Raises:
+        ValueError: if no receivers are provided.
```
daniel-zullo-frequenz: I'd be more precise here as it was done for `_Merge.__init__()`, otherwise it sounds to me like one receiver might be enough

```suggestion
        ValueError: if less than two receivers are provided.
```

src/frequenz/channels/_merge.py
```diff
@@ -29,25 +39,34 @@ class Merge(Receiver[_T]):
...
+class _Merge(Receiver[_T]):
+    """A receiver that merges messages coming from multiple receivers into a single stream."""
+
+    def __init__(self, *receivers: Receiver[_T], name: str | None) -> None:
+        """Create a `_Merge` instance.
 
         Args:
-            *args: sequence of channel receivers.
+            *receivers: The receivers to merge.
+            name: The name of the receiver. Used to to create the string representation
```
daniel-zullo-frequenz:
```suggestion
            name: The name of the receiver. Used to create the string representation
```

daniel-zullo-frequenz (commented):
I only found a few cosmetics in the documentation to check for, otherwise LGTM

src/frequenz/channels/_merge.py
```diff
@@ -13,13 +13,23 @@
...
-    """Merge messages coming from multiple channels into a single stream.
+def merge(*receivers: Receiver[_T]) -> Receiver[_T]:  # noqa: DOC502
+    """Merge messages coming from multiple receivers into a single stream.
+
+    Args:
+        *receivers: The receivers to merge.
+
+    Returns:
+        A receiver that merges the messages coming from multiple receivers into a
+            single stream.
```
daniel-zullo-frequenz: `The receiver` instead according to dart suggestions?

"The" also for arguments and return documentation. The rationale is that for attributes/arguments/returns you are talking about a specific instance ("this" instance), so a definite article is more appropriate, while for classes you are talking about an unspecified instance, any instance, so an indefinite article is more appropriate.

src/frequenz/channels/_merge.py
llucax: Ah, yeah, I forgot to update that docstring too, at first I checked that it wasn't empty, but then though it doesn't make much sense for only one receiver either.

src/frequenz/channels/_merge.py
llucax: Hahaha, well played Zullo, well played... (BTW, the rationale comes from me, not the dart docs, so you are sending me a citation of what I wrote myself :rofl: )

BTW, I thought about A/the here, and since it is not returning an attribute, but creating a complete new object, so every call will return a new one, not "the" same one, I thought "A" was more appropriate in this context. I only take these guides as suggestion, not law, and chose whatever I feel is more appropriate for each case.

Let me know if you still feel it should be updated in this case.

daniel-zullo-frequenz: haha I thought it was taken from the dart suggestions :angel: 

That's fine with me as you have thought about it and it wasn't randomly documented. I agree those are just suggestions and there is no need to be strict with them. Though I'd prefer to keep documentation consistent whenever is possible. So far, "The" seems to be always a good fit for arguments.

daniel-zullo-frequenz (dismissed):

src/frequenz/channels/_merge.py
```diff
@@ -53,7 +56,13 @@ def __init__(self, *receivers: Receiver[_T], name: str | None) -> None:
             *receivers: The receivers to merge.
             name: The name of the receiver. Used to to create the string representation
                 of the receiver.
+
+        Raises:
+            ValueError: if less than two receivers are provided.
         """
+        if len(receivers) < 2:
+            raise ValueError("At least two receivers must be provided")
```
shsms: if it is a single component, we could actually just return the receiver, instead of a `_Merge`.  and raise only when there are 0 receivers.

that way, users can still say "get receivers for all batteries and  merge them" and have to do special case handling only for one case.

shsms (commented):
Sorry about the late review, just got one comment.

src/frequenz/channels/_merge.py
llucax: True. That said, now I wonder if we shouldn't allow also empty `merges()` just for the sake of generality, so in general if people are doing something like:

```py
receivers = []
if x:
    receivers.add(x)
if y:
    receivers.add(x)
async for m in merge(*receivers):
    ...
```

Without needing to special case.

Is not exactly the same as when there is only one receiver, as if there are no receivers, the `async for` loop will end without any iterations, so I'm not 100% convinced which would be the most useful default, as users will probably need to handle the empty case specially in any case... :thinking:

shsms: I agree, maybe all we need are some logs instead of exceptions.

Marenz: We could have a c'tor flag or parameter `allow_empty=` or `min_required_receivers =`

llucax: I'm also not convinced about the logging, as if you call `merge()` with no receivers (or one) on purpose, then you'll get your logs littered with basically noise. We could add those ctor flags to decide if we raise or emit logs but I find it overkill to be honest.

My feeling for now is:

* `merge(receiver)` should be fine, there are really no surprises there, raising or logging here is being overly patronizing, as everything will work as expected.
* `merge()` should raise. It needs special casing anyway, as it would be very weird that one wants the `merge()` loop to exit without doing any iteration, and if that's your intention, then it can be explicitly expressed using:

  ```python
  receivers = []
  if x:
      receivers.add(x)
  if y:
      receivers.add(x)
  if receivers:
      async for m in merge(*receivers):
          ...
  ```

  Which is only one extra line and makes everything very clear and explicit.

llucax: OK, I went with this approach (raise with no receivers, optimize to return the same receiver if only one is provided). I also moved the raising to the `merge()` function, as `_Merge` is a low level construct. This way I can also avoid the `# noqa` for the indirectly raised exception.

llucax:
Updated. Sorry @shsms, I forgot to create fixup/new commits for all the changes and amended one one the old commits (to only raise when no receivers are passed). The diff for that amend is [this one](https://github.com/frequenz-floss/frequenz-channels-python/compare/d67b4d63b916efd494803ba0a1c8fe6bc5788416..f9c076f0d86f4fe2765961dabc6f4e4f9521044e).

All the other changes are in new commits.

shsms:
Thanks @llucax, I have no further comments, will approve once you fixup.

llucax:
Rebased to apply the fixups and enabled auto-merge.

shsms (approved):

PR #241, opened 2023-11-21
Fix timer `reset()` while it is being waited on

llucax:
If the timer was reset while it was being waited on, the next tick time was not recalculated and the timer would wait for the original time instead of the new one.

This means a timeout could fire even if there was actually no timeout (because the timer was reset).

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -145,4 +145,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* `Timer`: Fix bug that was causing calls to `reset()` to not reset the timer, if the timer was already being awaited.
--- a/src/frequenz/channels/timer.py
+++ b/src/frequenz/channels/timer.py
@@ -696,10 +696,15 @@ class Timer(Receiver[timedelta]):
 
         now = self._now()
         time_to_next_tick = self._next_tick_time - now
+
         # If we didn't reach the tick yet, sleep until we do.
-        if time_to_next_tick > 0:
+        # We need to do this in a loop also reacting to the reset event, as the timer
+        # could be reset while we are sleeping, in which case we need to recalculated
+        # the time to the next tick and try again.
+        while time_to_next_tick > 0:
             await asyncio.sleep(time_to_next_tick / 1_000_000)
             now = self._now()
+            time_to_next_tick = self._next_tick_time - now
 
         # If a stop was explicitly requested during the sleep, we bail out.
         if self._stopped:
--- a/tests/test_timer_integration.py
+++ b/tests/test_timer_integration.py
@@ -1,0 +1,34 @@
+# License: MIT
+# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
+
+"""Integration tests for the timer."""
+
+
+import asyncio
+from datetime import timedelta
+
+import async_solipsism
+import pytest
+
+from frequenz.channels.timer import Timer
+
+
+@pytest.mark.integration
+async def test_timer_timeout_reset(
+    event_loop: async_solipsism.EventLoop,  # pylint: disable=redefined-outer-name
+) -> None:
+    """Test that the receiving is properly adjusted after a reset."""
+
+    async def timer_wait(timer: Timer) -> None:
+        await timer.receive()
+
+    async with asyncio.timeout(2.0):
+        async with asyncio.TaskGroup() as task_group:
+            timer = Timer.timeout(timedelta(seconds=1.0))
+            start_time = event_loop.time()
+            task_group.create_task(timer_wait(timer))
+            await asyncio.sleep(0.5)
+            timer.reset()
+
+    run_time = event_loop.time() - start_time
+    assert run_time >= 1.5


llucax:
Based on #240.

The test is flaky, I will update it to remove the flakiness.

llucax:
Updated, hopefully this time it passes all the tests.

src/frequenz/channels/timer.py
```diff
@@ -696,10 +696,15 @@ async def ready(self) -> bool:  # noqa: DOC502
 
         now = self._now()
         time_to_next_tick = self._next_tick_time - now
+
         # If we didn't reach the tick yet, sleep until we do.
-        if time_to_next_tick > 0:
+        # We need to do this in a look also reacting to the reset event, as the timer
+        # could be reset while we are sleeping, in which case we need to recalculated
+        # the time to the next tick and try again.
```
daniel-zullo-frequenz:
```suggestion
        # We need to do this in a loop also reacting to the reset event, as the timer
        # could be reset while we are sleeping, in which case we need to recalculate
        # the time to the next tick and try again.
```

llucax:
#240 merged, rebased.

llucax:
Enabled auto-merge, please don't approve if you have suggestions.

tests/test_timer_integration.py
```diff
@@ -0,0 +1,31 @@
...
+    async with asyncio.timeout(2.0):
+        async with asyncio.TaskGroup() as task_group:
+            timer = Timer.timeout(timedelta(seconds=1.0))
+            start_time = event_loop.time()
+            task_group.create_task(timer_wait(timer))
+            await asyncio.sleep(0.5)
+            timer.reset()
+
+    run_time = event_loop.time() - start_time
+    assert run_time >= 1.5
```
daniel-zullo-frequenz, line 34: I'd also check for run_time < 2.0
```suggestion
    assert 1.5 >= run_time < 2.0
```

daniel-zullo-frequenz (commented):
I have a couple of suggestions to check for. LGTM otherwise

RELEASE_NOTES.md
```diff
@@ -145,4 +145,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+* `Timer`: Fix bug where if `reset()` was called while the timer was being waited on, the timer would not be reset and still wait for the original timeout.
```
daniel-zullo-frequenz: I'd suggest either `where` or `if` but not both
```suggestion
* `Timer`: Fix bug if `reset()` was called while the timer was being waited on, the timer would not be reset and still wait for the original timeout.
```

llucax: That was actually written by copilot, I thought it could do English better than me :P

ChatGPT says:

image

All robots seem to be on my side :laughing: but let me know if you still believe the `where` is incorrect. Of course I trust you more than the robots...

tests/test_timer_integration.py
llucax: Well, the whole block is inside a `with asyncio.timeout(2)`, so it will never be `> 2.0`, as in that case the `async with` block it will raise an exception.

src/frequenz/channels/timer.py
llucax: Amended :+1:

RELEASE_NOTES.md
daniel-zullo-frequenz: I'm afraid the robot is right :disappointed: 
My suggestion sounded like fixing the bug is conditional

tests/test_timer_integration.py
daniel-zullo-frequenz: oh right, my comment does not apply

tests/test_timer_integration.py
```diff
@@ -0,0 +1,31 @@
+"""Integration tests for the timer."""
```
daniel-zullo-frequenz, line 4: Do we need to add the license/copyright text here or it isn't needed?

daniel-zullo-frequenz (commented):
Just have a question about the missing copyright/license header in the test

RELEASE_NOTES.md
shsms: "fix bug where if" sounds more accurate, but it might be clearer to reorder the clauses, to keep the two words apart: "where x was failing if done while y was going on"

Fix bug that was causing calls to `reset` to not reset the timer, if the timer was already being awaited.

tests/test_timer_integration.py
llucax: Not sure if it is legally needed but we are doing it, will add.

llucax: Done!

RELEASE_NOTES.md
llucax: OK, the robot still agrees with me (which actually means agreeing with the other robot).

image

But I amended the commit to use @shsms proposed phrasing anyway, as it seems the robots favorite was not clear enough for humans :smile:

tests/test_timer_integration.py
```diff
@@ -0,0 +1,34 @@
+# License: MIT
+# Copyright © 2022 Frequenz Energy-as-a-Service GmbH
```
daniel-zullo-frequenz:
```suggestion
# Copyright © 2023 Frequenz Energy-as-a-Service GmbH
```

llucax: Fixed

daniel-zullo-frequenz (approved):

PR #243, opened 2023-11-22
Make `Merge` public again and rename it to `Merger`

llucax:
We need to make `Merge` public again, because the `Merge` receiver has more methods than a plain `Receiver`, in particular `stop()`, which users need to use for guaranteed cleanup.

But in this case it is better to introduce a breaking change to make sure users update to using the new `merge()` function, so we rename it to `Merger`. Also using a noun for the name is more in line with the other classes.

This also means now the `merge()` function returns the `Merger` instance instead of an abstract `Receiver`.

Additionally there are a couple of small unrelated cleanups:

* Remove `darglint`-related stuff, it was replaced by `pydoclint` a while ago.
* Remove cookiecutter template TODO comment.


llucax:
Enabling auto-merge

src/frequenz/channels/_merge.py
```diff
@@ -13,7 +15,7 @@
 _T = TypeVar("_T")
 
 
-def merge(*receivers: Receiver[_T]) -> Receiver[_T]:
+def merge(*receivers: Receiver[_T]) -> Merger[_T]:
```
daniel-zullo-frequenz, line 18: The `merge()` function returns the `Merge` instance instead of a `Receiver`. Should this be mentioned in the release notes?

daniel-zullo-frequenz (commented):
I only have a question and LGTM otherwise

src/frequenz/channels/_merge.py
llucax: No, because `merge()` is unreleased so far.

llucax:
@shsms last opportunity to complain, if @daniel-zullo-frequenz approves I will merge if you didn't complain before :laughing:

daniel-zullo-frequenz (approved):

PR #244, opened 2023-11-23
Add some useful mkdocs extensions

llucax:
We also make some initial use of them, but this is also in preparation of the creation of the user guide.

- Add the `mkdocs-include-markdown-plugin`
- Add the `markdown-svgbob` extension
- Render `Timer` diagrams with svgbob
- Reuse parts of `README.md` in `docs/index.md`

### DIFF
--- a/src/frequenz/channels/timer.py
+++ b/src/frequenz/channels/timer.py
@@ -87,12 +87,16 @@ class TriggerAllMissed(MissedTickPolicy):
         `T5`, happens at 5.1 (0.1 seconds late), so it triggers immediately
         again. The seventh tick, `T6`, happens at 6.0, right on time.
 
-        ```
-        0         1         2         3         4  o      5         6
-        o---------|-o-------|--o------|---------|--o------|o--------o-----> time
+        <center>
+        ```bob
+        0         1         2         3         4   T4    5         6
+        *---------o-*-------o--*------o---------o--**-----o*--------*-----> time
         T0          T1         T2                  T3      T5       T6
-                                                   T4
+
+        -o- "Expected ticks"
+        -*- "Delivered ticks"
         ```
+        </center>
     """
 
     def calculate_next_tick_time(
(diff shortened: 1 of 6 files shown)


llucax:
Based on #243, only the last 4 commits are for this PR.

llucax:
BTW, this is an example of how the diagrams are rendered:

image

src/frequenz/channels/timer.py
```diff
@@ -87,12 +87,16 @@ class TriggerAllMissed(MissedTickPolicy):
         `T5`, happens at 5.1 (0.1 seconds late), so it triggers immediately
         again. The seventh tick, `T6`, happens at 6.0, right on time.
 
-        ```
-        0         1         2         3         4  o      5         6
-        o---------|-o-------|--o------|---------|--o------|o--------o-----> time
+        <center>
+        ```bob
+        0         1         2         3         4   T4    5         6
+        *---------o-*-------o--*------o---------o--**-----o*--------*-----> time
```
daniel-zullo-frequenz, line 93: There is a double `delivered ticks` for t4 now, is this correct?

daniel-zullo-frequenz: ~also at 6 should be `o` instead of `*`~

src/frequenz/channels/timer.py
```diff
@@ -187,11 +197,17 @@ class SkipMissedAndDrift(MissedTickPolicy):
         so is right on time (no drift) and the same happens for tick `T6.3`,
         which triggers at 6.3 seconds.
 
-        ```
+        <center>
+        ```bob
         0         1         2         3         4         5         6
-        o---------|-o-------|--o------|---------|--o------|--o------|--o--> time
-        T0          T1         T2.2                T3.2      T5.3      T6.3
+        *---------o-*-------|-o*------|-O-------|-o*------|--*------|--*--> time
```
daniel-zullo-frequenz, line 203: this is still using `|`

daniel-zullo-frequenz (commented):
I have a couple of questions related to the new diagrams. LGTM otherwise

src/frequenz/channels/timer.py
llucax: Oh, good catch! Will fix.

src/frequenz/channels/timer.py
llucax: Yes.

> The fourth tick, `T3`, happens at time 4.3 (1.3 seconds late), so it also triggers immediately as well as the fifth tick, `T4`, which was also already delayed (by 0.3 seconds), so it catches up.

Before this was graphed by putting both ticks at the same time, like:

```
 4  o 
-|--o--
    T3
    T4
```

I think it is more clear to put them sequentially, one right after the other, as this is how it really happens, and one can see the small burst more clearly.

src/frequenz/channels/timer.py
llucax: Oh, no, wait, I left those on purpose, just to mark the seconds units. In the other graphs, all the expected ticks are in exactly the second boundary (because there is no drift), this is why there is no `|` (because I couldn't find any way to combine `|` and `o` in svgbob, otherwise that would have been the ideal). But in this case that there is drift, I think it is good to also show more clearly the second boundaries in the graph.

You can even see the screenshot of how is it rendered in a comment above.

If you think it is not clear, then we could discuss how to change it, but it was left as it is intentionally.

llucax:
Rebased on the current v1.x.x.

src/frequenz/channels/timer.py
daniel-zullo-frequenz: Ok, I understand now. The screenshot looks good

daniel-zullo-frequenz (approved):

PR #245, opened 2023-11-24
Add a User Guide

### LINKED ISSUE #177: Write introductory documentation
### What's needed?

We need a documentation to introduce new users to channels.

### Proposed solution

Write a document that explain the basic channels concepts:

* Channels
* Senders
* Receivers
* ?

It should also mention common pitfalls, like messages should be small and serializable even when there is no networking support yet.

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
This PR introduces a new `User Guide` top level section in the documentation explaining how to use the library in a more practical and cohesive way than just reading through the API documentation.

The bulk of the documentation is still hosted in the code itself, as docstrings, so users can still access most of user guide content directly from their IDE.

It also improves the README and the docs home page, adding a quick start section to both.

Fixes #177.


llucax:
This is based on:

* #244

Only the commits not present in that PR should be reviewed.

llucax:
You can see the results here: https://llucax.github.io/frequenz-channels-python/v1.0-pre/user-guide/quick-start/

A few comments about stuff I found or thought about while doing this guide:

* Having a higher level documentation in the module documentation makes a lot of sense IMHO, and it is a shame that this documentation is not available in the IDE, so I would like to re-consider the module structure to find a way to expose it. We can discuss elsewhere to avoid polluting this PR too much.
* Having a `channel` module with a `Channel` abstract class/protocol, would also allow us to expose docs about general channel concepts that is available in IDEs too.

llucax:
Rebased

README.md
```diff
@@ -28,6 +28,11 @@ channels](https://tour.golang.org/concurrency/2) but it also borrows ideas from
 
 <!-- supported-platforms -->
 
+!!! Note inline end
+    Other platforms are likely to work too, but they are not
+    automatically tested, so we cannot guarantee that they will
+    work.
```
shsms: I wonder if we have a strong requirement for Python 3.11 in channels.  We use type guards, but those are there in 3.10.

README.md
```diff
@@ -61,12 +65,189 @@ Then, just install using `pip`:
...
+
+            await asyncio.sleep(5.0)
+
+            # Stop senders, after 2 seconds not receiving any data,
+            # the receiver will stop too
+            await control_sender.send(Command.STOP_SENDER)
+
+
+    asyncio.run(main())
+    ```
```
shsms: Putting long examples in the readme feels a bit dangerous, because they won't be CI tested, and we can't verify correctness with just a glance.  Unless I'm wrong about CI checking the readme examples, I'd rather have a couple of small examples here,  and link to source files for the longer examples.

src/frequenz/channels/_anycast.py
```diff
@@ -20,54 +20,165 @@
...
-    Anycast channels support multiple senders and multiple receivers.  A message sent
-    through a sender will be received by exactly one receiver.
+    """A channel that delivers each message to exactly one receiver.
+
+    # Description
+
+    [Anycast][frequenz.channels.Anycast] channels support multiple
+    [senders][frequenz.channels.Sender] and multiple
+    [receivers][frequenz.channels.Receiver]. Each message sent through any of the
+    senders will be received by exactly one receiver.
```
shsms: Might be worth mentioning, but I'll let you decide:

- The first receiver that is awaited will get the next message.  When multiple receivers are waiting, the scheduler picks a receiver for each next massage.
- Anycast channels behave exactly like the channels in Golang.

src/frequenz/channels/_anycast.py
```diff
@@ -20,54 +20,165 @@
...
+
+    <center>
+    ```bob
+    .---------. msg1                           msg1  .-----------.
+    | Sender  +------.                       .------>| Receiver  |
+    '---------'      |      .----------.     |       '-----------'
+                     +----->| Channel  +-----+
+    .---------.      |      '----------'     |       .-----------.
+    | Sender  +------'                       '------>| Receiver  |
+    '---------' msg2                           msg2  '-----------'
```
shsms, line 45: This could be a bit confusing.  Maybe we specify that this is the behaviour when both receivers are awaited?

Because it is also possible that both messages go to one receiver. When one of the workers is slower, etc.

src/frequenz/channels/_merge.py
```diff
@@ -1,7 +1,51 @@
...
+
+async for value in merge(receiver1, receiver2):
+    print(value)
+```
+
+If the first message comes from `channel2` and the second message from `channel1`, the
+first message will be received immediately, and the second message will be received as
+soon as it is available.
+
+This is extremely convenient when you just need to receive messages and don't care about
```
shsms: extremely sounds too much like an opinion.  If I was doing it, I would say "This can/might be useful when ...", etc.

But maybe "very convenient" is a middle point?

src/frequenz/channels/_receiver.py
```diff
@@ -16,7 +145,7 @@
 
 
 class Receiver(ABC, Generic[_T]):
-    """A channel Receiver."""
+    """An entity that receives values from a channel."""
```
shsms: I guess we need to suggest that we receive the values, but through these objects.  

Maybe something like this?

```suggestion
    """An type for receiving values from a channel."""

File: src/frequenz/channels/_select.py
Comment (shsms) [inline src/frequenz/channels/_select.py:1]: > This module is not publicly available, so users won't be able to access
> the information in IDEs for example, but it will be rendered as part of
> the User Guide.

I don't remember if you already explained what this means, but maybe you could show a demo in a call so it is clear to everyone.

File: src/frequenz/channels/_select.py [hunk 7]
```diff
@@ -3,9 +3,137 @@
 
 """Select the first among multiple Receivers.
 
-Expects Receiver class to raise `StopAsyncIteration`
-exception once no more messages are expected or the channel
-is closed in case of `Receiver` class.
+# Usage
+
+If you need to receiver different types of messages from different receivers, you need
```
shsms, line 8:
```suggestion
If you need to receive different types of messages from different receivers, you need

File: src/frequenz/channels/_select.py [hunk 8]
```diff
@@ -3,9 +3,137 @@
...
+to know the source of a particular received value to know the type of the value.
+
+[`select()`][frequenz.channels.select] allows you to do that. It is an
+[async iterator][typing.AsyncIterator] that will iterate over the values of all
+receivers as they receive new values.
+
+It yields a [`Selected`][frequenz.channels.Selected] object that will tell you the
+source of the received message. To make sure the received value is *casted* to the
```
shsms: Quoting from: https://www.wordreference.com/englishusage/cast
> The verb cast has several other meanings. Note that for all its meanings its past tense and past participle is cast, not ‘casted’.

```suggestion
source of the received message. To make sure the received value is *cast* to the

**Review by shsms — COMMENTED** [review]
I've reviewed upto the `select` commit.  Will continue with the rest soon.

File: README.md (hunk 1 above)
Comment (llucax) [reply README.md:35]: We are using exception groups.

Comment (llucax) [reply README.md]: We are using exception groups.

File: README.md (hunk 2 above)
Comment (llucax) [reply README.md]: Yeah, this is an issue with all the documentation in `docs/`, but we should start testing all docs in the CI soon, I don't think the solution is to add smaller examples. Now that we start having more complete docs and we'll need to main them I think it makes sense to prioritize extending the CI testing soon rather that avoid examples in docs.

File: src/frequenz/channels/_anycast.py (hunk 3 above)
Comment (llucax) [reply src/frequenz/channels/_anycast.py]: Sounds good, I didn't know about channels in golang being anycast, so TIL.

File: src/frequenz/channels/_anycast.py (hunk 4 above)
Comment (llucax) [reply src/frequenz/channels/_anycast.py:45]: Yeah, I thought about that, but I think if we graph all the possible combinations it can get pretty messy. The idea of this diagram was to quickly and briefly show how the channel works conceptually. Maybe we should clarify this in more visible way though (like a *IMPORTANT* admonition or something), but I think making the diagram more complicated will not help much.

Comment (llucax) [reply src/frequenz/channels/_anycast.py:45]: What do you think about this?

```diff
     messages in the channel's buffer. The channel's buffer size can be configured at
     creation time via the `limit` argument.
 
+    The first receiver that is awaited will get the next message. When multiple
+    receivers are waiting, the [asyncio][] loop scheduler picks a receiver for each next
+    massage.
+
+    This means that, in practice, there might be only one receiver receiving all the
+    messages, depending on how tasks are schduled.
+
+    If you need to ensure some delivery policy (like round-robin or uniformly random),
+    then you will have to implement it yourself.
+
     To create a new [senders][frequenz.channels.Sender] and
     [receivers][frequenz.channels.Receiver] you can use the
```

src/frequenz/channels/_merge.py
llucax: Yeah, true, just "convenient" would do too. :smile:

llucax: I'm going with "This can be convenient when..."

src/frequenz/channels/_receiver.py
llucax: Mmm, not sure, in general I think "A class for...", "A function for...", "An enum that..." all seems redundant when you already know the type of symbol you are dealing with, and this looks a bit about that. Asking for ideas to ChatGPT it suggests "endpoint", which seems better to me, so what about "An endpoint for receiving messages."? (I wouldn't include the "channels" part either, as we have receivers that are not attached to channels)

src/frequenz/channels/_select.py
llucax: Sure. Maybe it is more confusing because it's not entirely correct, people can access the module, is just that because is not present in the docs, they probably will not even know about their existence. What I mean is:

```console
$ python -c 'import frequenz.channels; help(frequenz.channels)' | head -n10
Help on package frequenz.channels in frequenz:

NAME
    frequenz.channels - Frequenz Channels.

DESCRIPTION
    This package contains
    [channel](https://en.wikipedia.org/wiki/Channel_(programming)) implementations.
    
    Base classes:
```

This is also what's shown by the IDE when you have the help available:

image

What I mean is, if people don't know `frequenz.channels._select` exist, they will never do:

```console
$ python -c 'from frequenz.channels import _select; help(_select)' | head -n10
Help on module frequenz.channels._select in frequenz.channels:

NAME
    frequenz.channels._select - Select the first among multiple Receivers.

DESCRIPTION
    # Usage
    
    If you need to receiver different types of messages from different receivers, you need
    to know the source of a particular received value to know the type of the value.
```

Or:

image

To see `select()`'s high-level documentation.

I hope it is clearer now, but we can also clarify in a call or something.

README.md
shsms: Then maybe we can mention Python 3.11 as a min requirement.  Else it sounds like lower versions might work, when we know already know that's not the case.

20231206_16h18m27s_grim

README.md
shsms: I wasn't suggesting we avoid long examples, just that we import them from source files instead.

src/frequenz/channels/_anycast.py
shsms: > Maybe we should clarify this in more visible way though

That's all I was suggesting, as a description for the image.  Something like, "This depicts a case where two receivers are waiting for messages".

I guess the diff you pasted is supposed to be for a different comment? https://github.com/frequenz-floss/frequenz-channels-python/pull/245#discussion_r1417019182

But that looks good as well.

src/frequenz/channels/_merge.py
shsms: Another possibility: "This can be helpful when ...".

Also might be worth mentioning: "... when you don't care about where they are coming from, or if the messages themselves have an id field that would help identify their origin or context ..."

src/frequenz/channels/_receiver.py
shsms: "Interface for receiving messages from a stream"?

src/frequenz/channels/_select.py
shsms: Oh yes, thanks for explaining!

Now I remember discussing this with you, don't remember what the conclusion was.  I'm assuming it was to keep modules private as long as they export one thing, but use public modules when something exposes multiple features?

src/frequenz/channels/_sender.py
```diff
@@ -1,7 +1,53 @@
...
+from frequenz.channels import Anycast
+
+channel = Anycast[int](name="test-channel")
+sender = channel.new_sender()
+
+await sender.send("Hello, world!")
+```
+
+Even when [`send()`][frequenz.channels.Sender.send] is an asynchronous method, some
+channels may implement it as a synchronously. For example, buffered channels that
```
shsms:
```suggestion
Although [`send()`][frequenz.channels.Sender.send] is an asynchronous method, some
channels may implement it synchronously. For example, buffered channels that

Comment (flora-hofmann-frequenz) [inline src/frequenz/channels/_sender.py]: This is a bit confusing. Is it referring to asynchronously or synchronously?

**Review by flora-hofmann-frequenz — COMMENTED** [review]
Only a small comment from my side. Besides, I mostly agree with Sahas' additions (indicated by 'thumbs up').

File: src/frequenz/channels/_sender.py [hunk 10]
```diff
@@ -12,7 +58,7 @@
 
 
 class Sender(ABC, Generic[_T]):
-    """A channel Sender."""
+    """An entity that sends values to a *channel*."""
```
shsms: Here too, entity doesn't feel like a suitable word to describe an abc.

src/frequenz/channels/file_watcher.py
```diff
@@ -16,16 +31,16 @@
...
-    """A new file was created."""
+    """The file was created."""
 
     MODIFY = Change.modified
-    """An existing file was modified."""
+    """The file was modified."""
 
     DELETE = Change.deleted
-    """An existing file was deleted."""
+    """The file was deleted."""
```
shsms: We call it `FileWatcher`, but at least on linux, we are able to watch directories as well I think, although I don't know if we have tests, etc.  Maybe we should support that, but probably for a separate issue.

Btw "The types of changes that can be observed" can maybe change to "The types of file events that can be observed"?

src/frequenz/channels/file_watcher.py
```diff
@@ -34,12 +49,73 @@ class Event:
...
+            # Create file
+            with open(PATH, "w", encoding="utf-8") as file:
+                file.write("Hello, world!")
+            async with asyncio.TaskGroup() as group:
+                group.create_task(update_file())
+                async for selected in select(file_watcher):
+                    if selected_from(selected, file_watcher):
+                        event = selected.value
+                        print(f"File {event.path}: {event.type.name}")
+                        break
```
shsms: why select?

shsms (commented):
Some more comments,  but keeping the `timer` commit for the afternoon, because it is a bit big.  So the last 3 commits are pending.

src/frequenz/channels/timer.py
```diff
@@ -1,14 +1,90 @@
...
+                drift = selected.value
+                print(f"No data received for {timer.interval + drift} seconds, giving up")
+                break
+
+
+    asyncio.run(main())
+    ```
+
+    This timer will *rearm* itself automatically after it was triggered, so it will trigger
+    again after the selected interval, no matter what the current drift was.
```
shsms, line 61: So `timeout` is actually also `periodic`.  And calling it a `timeout` timer doesn't clarify this.

Because when there are missed ticks, even the thing we call `periodic`, can't behave periodically, it just sends a burst of missed ticks, which is totally not periodic.  Triggering missed ticks makes the data look periodic, but it doesn't make the timer periodic. I guess software timers just can't be truly periodic, especially in a co-operative multi-tasking context.

They are both equally periodic or non-periodic based on whether there are missed ticks.  Maybe it was a bad decision to give names to these, and we should have exposed just the missed tick policies instead.

Sorry about bringing this up all over again, just ranting.  Feel free to ignore or defer until next major version.

shsms: I see you've clarified these points below, but the names themselves aren't helping, I think.

src/frequenz/channels/timer.py
```diff
@@ -282,116 +426,60 @@ def __repr__(self) -> str:
...
+
+    Warning:
+        Even when the [`asyncio`][] loop's monotonic clock is a [`float`][], timers use
+        `int`s to represent time internally. The internal time is tracked in
+        microseconds, so the timer resolution is 1 microsecond
+        ([`interval`][frequenz.channels.timer.Timer.interval] must be at least
+         1 microsecond).
+
+        This is to avoid floating point errors when performing calculations with time,
+        which can lead to very hard to reproduce, and debug, issues.
```
shsms:
```suggestion
        which can lead to issues that are very hard to reproduce and debug.

**Review by shsms — COMMENTED** [review]
Finished reviewing.  This is amazing work, Luca!  very comprehensive.

File: src/frequenz/channels/file_watcher.py (hunk 12 above)
Comment (llucax) [reply src/frequenz/channels/file_watcher.py]: I think I copied it from the tests, and the tests had other stuff in the select, I can remove the select :+1:

File: README.md (hunk 1 above)
Comment (llucax) [reply README.md]: OK, I will try to clarify.

File: src/frequenz/channels/_anycast.py (hunk 4 above)
Comment (llucax) [reply src/frequenz/channels/_anycast.py:45]: Yeah, sorry

Comment (llucax) [reply src/frequenz/channels/_anycast.py:45]: About adding the caption, I didn't find a nice way to do it. [mkdocs-material captions](https://squidfunk.github.io/mkdocs-material/reference/images/#image-captions) seemed like a perfect fit, but having a *svgbob* diagram instead of an image is not supported (the code is not converted to an image) so it looks really ugly.

Then I tried with:

```markdown
    <center style="clear: right">
    ```bob
    .---------. msg1                           msg1  .-----------.
    | Sender  +------.                       .------>| Receiver  |
    '---------'      |      .----------.     |       '-----------'
                     +----->| Channel  +-----+
    .---------.      |      '----------'     |       .-----------.
    | Sender  +------'                       '------>| Receiver  |
    '---------' msg2                           msg2  '-----------'

    ```

    <span style="font-size: smaller">*Two receivers waiting for messages sent by two
    senders*</span>
    </center>
```

![image](https://github.com/frequenz-floss/frequenz-channels-python/assets/1031485/e8246d1c-a708-4f22-916f-94455c8a7b18)

Which looks nice, but adds a lot of syntactic noise to the text docs.

Finally I tried:

```markdown
    <center>
    ```bob
    .---------. msg1                           msg1  .-----------.
    | Sender  +------.                       .------>| Receiver  |
    '---------'      |      .----------.     |       '-----------'
                     +----->| Channel  +-----+
    .---------.      |      '----------'     |       .-----------.
    | Sender  +------'                       '------>| Receiver  |
    '---------' msg2                           msg2  '-----------'

       Two receivers waiting for messages sent by two senders
    ```
    </center>
```

![image](https://github.com/frequenz-floss/frequenz-channels-python/assets/1031485/18d26cd2-1d1e-4481-9cd0-e253a689964e)

Which is much more lightweight, but then it doesn't look that nice.

My impression is adding the caption doesn't really add much more info, what it says seems pretty obvious from the diagram, and if it is not, maybe I should just remove the diagram, the idea was the other way around, that the diagram helps to understand the text not that we need text to understand the diagram :laughing: 

If you really think this needs clarification or really helps to understand better, then I can go with either of those 2 alternatives, I leave that to you.

src/frequenz/channels/_merge.py
llucax: I find the last a bit too specific of a use case, and doesn't fit very nice to the following part talking about select, I would need to rephrase the whole thing only to include this very specific example, so I will keep that as it is for now. If you feel it is worth mentioning maybe you can provide the full rephrased paragraph?

src/frequenz/channels/_receiver.py
llucax: Yeah, that was another suggestion by ChatGPT, but I still find it too much like class/enum/etc, so I preferred endpoint. But if you prefer it, I could go with it.

src/frequenz/channels/_select.py
llucax: If you are referring to the discussion about restructuring the channels modules, then we decided to only put `timer`, `file_watcher` and `event` in a separate module because they were more satellites. Otherwise I don´t remember talking about this specifically.

src/frequenz/channels/_sender.py
llucax: This is poorly worded, I guess the last "as a" shouldn't be there. Is this better?

> Even when [`send()`][frequenz.channels.Sender.send] is an asynchronous method, some
> channels may implement it synchronously.

Does that makes it clearer?

llucax: That's probably not much better, I think there is extra confusion because it is not mentioning blocking, which is the main point. What about this?

> Although [`send()`][frequenz.channels.Sender.send] is an asynchronous method, some
> channels may implement it in a synchronous, non-blocking way.

llucax: See https://github.com/frequenz-floss/frequenz-channels-python/pull/245#discussion_r1422611922

src/frequenz/channels/_sender.py
llucax: Yeah, I'm updating both to be consistent.

src/frequenz/channels/file_watcher.py
llucax: Well, depends on how pedantic we want to be, but technically in Unix a directory is a type of file, same as you have block devices files, etc. a directory is just a file with an index of other names and which inodes it points to.  I wonder if we shouldn't be more transparent about this being a wrapper on `awatch` and then relay all the specifics to the dependency... But yeah, for another issue.

I'll use "The types of file events that can be observed" :+1:

src/frequenz/channels/timer.py
llucax: We decided to remove `periodic()` and `timeout()` as we'll never find a name that can convey all the intricacies of timers in the async world. Users will have to pass the missing ticks policies manually. I will update the docs when I do that.

I'll probably leave the docs about both methods as examples of how to use the timer to achieve different goals.

llucax: * https://github.com/frequenz-floss/frequenz-channels-python/issues/253

llucax:
Updated with all the suggestions, except for the ones I left follow-up comments.

llucax:
Please mark the conversation that you happy with the resolution as resolved.

README.md
llucax: OK, in this case I don't see where this long example could be a good fit in the sources, so I rather leave it here and try to push for the testing of markdown files soon.

src/frequenz/channels/_anycast.py
shsms: That's true, and the behaviour is explained clearly in the text anyway, so I'm fine with leaving it as it is.

src/frequenz/channels/_merge.py
shsms: No, you're right.  It is probably already obvious that we're only talking about whether identity is available from the infrastructure.

src/frequenz/channels/_select.py
shsms: looks like this one is pending.

src/frequenz/channels/_sender.py
shsms: synchronous + non-blocking can be a bit confusing, but if people read further, I think it should become clearer.

src/frequenz/channels/_select.py
llucax: Oh, thanks for checking, it must have been lost in some rebase because I'm sure I updated the file at some point. Updated now.

shsms (commented):
Thanks luca, all looks good to me.  I will approve once you fixup.

llucax:
Fixup done, enabling auto-merge.

shsms (approved):

PR #264, opened 2024-01-16
`Timer`: Remove `periodic()` and `timeout()`

### LINKED ISSUE #253: `Timer`: Remove `periodic()` and `timeout()`
### What's needed?

The names are just too confusing and we'll never find a name that can convey all the intricacies of timers in the async world.

I'll probably leave the docs about both methods as examples of how to use the timer to achieve different goals.

### Proposed solution

Remove `periodic()` and `timeout()` and force users to pass the missing ticks policies manually.

llucax:
The names are just too confusing and we'll never find a name that can convey all the intricacies of timers in the async world, so it is better to just remove `periodic()` and `timeout()` and force users to pass the missing ticks policies manually.

Fixes #253.


src/frequenz/channels/timer.py
```diff
@@ -5,6 +5,13 @@
 
 # Quick Start
 
+Info: Important
+    This quick start is provided to have a quick feeling of how to use this module, but
+    it is extremely important to understand how timers behave when they are delayed.
+
+    We recommned emphatically to read about [missed ticks and
```
Marenz: recommned > recommend

tests/test_file_watcher_integration.py
```diff
@@ -27,7 +27,7 @@ async def test_file_watcher(tmp_path: pathlib.Path) -> None:
     expected_number_of_writes = 3
 
     file_watcher = FileWatcher(paths=[str(tmp_path)])
-    timer = Timer.timeout(timedelta(seconds=0.1))
+    timer = Timer(timedelta(seconds=0.1), SkipMissedAndDrift())
```
shsms, line 30: do you think `SkipMissedAndDrift` should be an instance created in the channel package, so that we don't have to specify the `()` to create an instance of the thing?

This was a bit confusing for me the first time I did this, because I was expecting to pass just the policy, not an object of the policy that I had to create.

Marenz (changes requested):
Minor typo

shsms:
Although not strictly necessary, in this case it is easy to make all commits pass CI by just moving the first commit to near the end.

tests/test_file_watcher_integration.py
llucax: The policies can have options, and `SkipMissedAndDrift()` have one ([`delay_tolerance`](https://frequenz-floss.github.io/frequenz-channels-python/v1.0-dev/reference/frequenz/channels/timer/#frequenz.channels.timer.SkipMissedAndDrift.delay_tolerance)). I think if we provide fixed instances with fixed options we'll eventually need another PR to remove them because they are still hiding the complexity :)

llucax:
> Although not strictly necessary, in this case it is easy to make all commits pass CI by just moving the first commit to near the end.

True, but I thought review-wise It makes more sense to start with the important change first, and then adapt the code to deal with that change. I'm trying to prioritize easy of review when splitting the commits.

I can move the commit if you prefer though...

src/frequenz/channels/timer.py
llucax: Fixed

llucax:
Updated to fix the typo and move the removal from the code to the end.

llucax:
Enabled auto-merging.

Marenz (approved):

Issue #239, opened 2023-11-16
`Receiver.map()` doesn't preserve the original receiver's type


llucax:
### What happened?

When calling `map()` on a receiver, the returned type is of type `_Map` and tying-wise it is an abstract `Receiver`, this means that when mapping a receiver that has more methods, like `.close()` for merging or `.reset()` for `Timer`, these methods are not available anymore after `.map()` is used.

### What did you expect instead?

I expect the type to be preserved, so the mapped receiver can be used the same way as the original receiver.

### Extra information

The only way to fix this is probably to have some mapping logic inside `Receiver` itself instead of returning a new type. That or removing `.map()` and just let users do the transformation themselves when consuming from the receiver.

llucax:
I don't think this is doable. Currently `map()` is defined as:

```py
def map(self, call: Callable[[_T_co], _U_co]) -> Receiver[_U_co]:
```

Where `_T_co` is the type parameter of the `Receiver` that defines this `map()`.

To be able to return a type that has all the methods of the receiver subclass (like `Timer`), besides the implementation details on how we create such a wrapper, type-system-wise we should be able to do something like:

```py
def map(self, call: Callable[[_T_co], _U_co]) -> Self[_U_co]:
```

I.e. we would have to return the concrete receiver type (like `Timer`) but instead of being parametrized with `_T_co` as the value type, it should be `_U_co`. This doesn't seem to be supported by Python.

And even if it were, it is basically impossible for all types that don't even have a type parameter. For example, `Timer` is not parametrized, it just inherits from `Receiver[timedelta]`, so the mapper receiver should be somehow `Timer[_U_co]`, which doesn't make any sense as `Timer` doesn't have a type parameter.

An alternative would be to just return `Any`, but that is probably much worse than just asking the users to save the original receiver if they want to call any extra methods from it at some point.

I suggest improving the documentation to inform the user about this issue and just recommend them to save a reference to the original receiver if they need to do something else with it.

llucax:
Clarification done in #271. Closing as invalid.

shsms:
This might actually become possible once we have default generics: https://peps.python.org/pep-0696/

llucax:
How could that help?

shsms:
1. type return value with `Self`
2. add `__getattr__` in `_Map` that retrieves functions from underlying object.
3. Change `Timer` etc to have default generic argument `timedelta`, etc., so that users don't have to specify it.

[...]

PR #271, opened 2024-02-02
Add a tip about the returned receiver type for `map()`

llucax:
Recommend saving a reference to the original receiver if users need to access methods that are not part of the `Receiver` interface.

Clarifies the issue pointed out by #239.


llucax:
Enabled auto-merge.

shsms (approved):

PR #274, opened 2024-03-04
Use the new set of labels

llucax:
The labels have been updated to better reflect the current state of the project. Update all labels-related configuration to reflect this change.

Also for the labeler configuration add some missing files that were not being labeled at all.


llucax:
Once this is approved, I will update the project's labels.

.github/ISSUE_TEMPLATE/bug.yml
```diff
@@ -51,8 +51,9 @@ body:
         - Unit, integration and performance tests (part:tests)
         - Build script, CI, dependencies, etc. (part:tooling)
         - Channels, `Broadcast`, `Anycast`, etc. (part:channels)
-        - Select (part:select)
-        - Utility receivers, `Event`, `FileWatcher`, `Timer`, etc. (part:receivers)
+        - Core types (`Sender`, `Receiver`, exceptions, etc.) (part:core)
+        - Utilities (`Event`, `FileWatcher`, `Timer`, etc.) (part:utilities)
+        - Multiplexing (`select()`, `merge()`, etc.) (part:multiplexing)
```
shsms: Maybe `synchronization` is better, because it is standard, and because `multiplexing` is about preparing for transmission, whereas these tools are more general.

Marenz: How about `event handling`

Marenz: "Event Polling"
"I/O Scheduling"
"Concurrency Handling"
"Stream Coordination"

shsms: The reason I think "Synchronization" works is because these are in a way task synchronization mechanisms, like mutexes are.

They take multiple receivers that would otherwise be processed asynchronously and makes sure one message is processed at a time, even when multiple messages are waiting.

shsms: If we didn't have that need, to synchronize their processing, we ideally shouldn't have to use these tools.  Although we still use them as alternative to `LatestValueCache`, etc.

llucax: For last value cache we could actually just have one task per value I guess.

I'm not convinced about any of the options (including multiplexing), but I'm OK with synchronization too (I find Marenz suggestions a bit too far off). The user guide says "Multiple sources" so that might be another alternative (`multiple-sources`).

We could also just leave `select` and add `merge` :shrug:

shsms: To me it makes sense that this is "Synchronization of channels", but we can use "select" and "merge" as well.

llucax: I was thinking that *Synchronization of multiple sources* sounds good, so we can use `Synchronization of multiple sources (`select()`, `merge()`, etc.) (part:synchronization)`. And now I wonder if we should also update the user guide section to match this.

Something like:

```diff
diff --git a/docs/user-guide/receiving/multiple-sources/index.md b/docs/user-guide/receiving/multiple-sources/index.md
index c1620a1..8534ae8 100644
--- a/docs/user-guide/receiving/multiple-sources/index.md
+++ b/docs/user-guide/receiving/multiple-sources/index.md
@@ -1,8 +1,9 @@
-# Multiple Sources
+# Synchronization of Multiple Sources
 
-If you need to receive messages from multiple sources it can be complicated, as
-you probably want to get the first message of each receiver as soon as it is
-available. A naive approach like this will not work:
+If you need to receive messages from multiple sources in a synchronized way it
+could be difficult to, for example, receive the first message of each receiver
+as soon as it is available in one single task. A naive approach like this will
+not work:
 
 ```python
 receiver1: Receiver[int] = channel1.new_receiver()
```

image

shsms (dismissed):

.github/ISSUE_TEMPLATE/bug.yml
shsms: That looks good to me.

shsms (approved):

PR #275, opened 2024-03-04
Enable and use GitHub Alert syntax for admonitions

llucax:
This allows admonitions/callouts/alerts to be rendered properly both in GitHub and `mkdocs`. On the downside, GitHub syntax is less flexible than `mkdocs`, so we can't have the admonitions be rendered *inline end* any more.

Because of this we only use the GitHub syntax for the README, as it is more important that the README is rendered correctly in GitHub than having it look a bit nicer in `mkdocs`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -28,16 +28,15 @@ channels](https://tour.golang.org/concurrency/2) but it also borrows ideas from
 
 <!-- supported-platforms -->
 
-!!! Note inline end
-
-    Newer Python versions and other operating systems and architectures might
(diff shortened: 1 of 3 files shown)


llucax:
Before:

image

After:

image

Marenz (approved):

shsms:
I wonder what requirement is missing.

llucax:
I guess it got borked when I switched the merge queue from a "branch protection" to a "ruleset" #GitHub :unicorn:

PR #276, opened 2024-03-04
Remove boilerplate from examples in the user guide

llucax:
We filter out all the boilerplate (importing, declaring some variables) from the code shown in the user guide, as it is distracting for such a high-level introduction. This means more examples listed in modules have the boilerplate removed from the generated docs. We do this by adding the [pymdownx-superfence-filter-lines] markdown extension.

Examples in classes or functions, that usually show up in the API docs are kept as before, with all the code. This is so users wanting to dive deeper into the library can still copy&paste code snippets from these examples and expect them to mostly work.

[pymdownx-superfence-filter-lines]: https://github.com/frequenz-floss/pymdownx-superfence-filter-lines-python

### DIFF
--- a/mkdocs.yml
+++ b/mkdocs.yml
@@ -83,6 +83,10 @@ markdown_extensions:
         - name: mermaid
           class: mermaid
           format: !!python/name:pymdownx.superfences.fence_code_format
+        - name: "*"
+          class: "highlight"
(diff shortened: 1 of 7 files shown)


llucax:
Before:

image

After:

image

Marenz (approved):
pretty cool

PR #279, opened 2024-03-05
Improve documentation

llucax:
This PR includes a lot of small documentation improvements and fixes, trying to make the documentation more consistent and easier to read.

The commits are not particularly very well split, as many stuff was done in parallel and the splitting was done afterwards. Cleaning it up is very time demaning as it produces a lot of conflicts when rebasing.

PR #281, opened 2024-03-06
Rename "value" to "message" and other minor breaking changes

llucax:
This PR replaces the term "value" with "message" as "value" is too generic and used in many other contexts.

It also makes `Selected.was_stopped` a property and make some function taking only one obvious argument positional-only, to avoid having the argument name being part of the interface (we actually renamed those arguments too, but that's not included in the release notes as they are not part of the public API any more anyways).

We also do some other minor internal renaming and changing of arguments to positional-or-keyword-only.

### DIFF
--- a/src/frequenz/channels/_select.py
+++ b/src/frequenz/channels/_select.py
@@ -6,17 +6,17 @@
 # Usage
 
 If you need to receiver different types of messages from different receivers, you need
-to know the source of a particular received value to know the type of the value.
+to know the source of a particular received message to know the type of the message.
 
 [`select()`][frequenz.channels.select] allows you to do that. It is an
-[async iterator][typing.AsyncIterator] that will iterate over the values of all
-receivers as they receive new values.
+[async iterator][typing.AsyncIterator] that will iterate over the messages of all
+receivers as they receive new messages.
 
 It yields a [`Selected`][frequenz.channels.Selected] object that will tell you the
-source of the received message. To make sure the received value is *cast* to the
+source of the received message. To make sure the received message is *cast* to the
 correct type, you need to use the [`selected_from()`][frequenz.channels.selected_from]
 function to check the source of the message, and the
-[`value`][frequenz.channels.Selected.value] attribute to access the message:
+[`message`][frequenz.channels.Selected.message] attribute to access the message:
 
 ```python show_lines="8:"
 from frequenz.channels import Anycast, ReceiverStoppedError, select, selected_from
@@ -28,9 +28,9 @@ receiver2 = channel2.new_receiver()
 
 async for selected in select(receiver1, receiver2):
     if selected_from(selected, receiver1):
-        print(f"Received from receiver1, next number: {selected.value + 1}")
+        print(f"Received from receiver1, next number: {selected.message + 1}")
     elif selected_from(selected, receiver2):
-        print(f"Received from receiver2, length: {len(selected.value)}")
+        print(f"Received from receiver2, length: {len(selected.message)}")
     else:
         assert False, "Unknown source, this should never happen"
 ```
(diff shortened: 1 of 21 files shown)


llucax:
Based on #279.

src/frequenz/channels/_select.py
```diff
@@ -228,16 +228,9 @@ def exception(self) -> Exception | None:
         """
         return self._exception
 
+    @property
     def was_stopped(self) -> bool:
```
shsms, line 232: `Selected.was_stopped` feels a bit ambiguous to me.  Might be worth renaming to `Selected.receiver_stopped`.

shsms (approved):
Just one comment.  Approving anyway because it might not be so bad to leave as it is.

src/frequenz/channels/_select.py
llucax: I find it clear enough but I'm also OK to rename it. Will wait a bit in case anyone else wants to chime in.

daniel-zullo-frequenz: I agree the suggested name is more precise. Even though  the new documentation will clear up any ambiguity I'd also suggest to rename it

llucax: OK, I was doing the change but then wondered what about `exception` and `message`, should they also be `receiver_exception` and `receiver_message`? I'm not super convinced about introducing an inconsistency where some stuff in `Selected` has a `receiver_` prefix and some not, but also find a bit overkill to prepend everything with `receiver_` as it is pretty clear to me that `Selected` is wrapping a receiver.

llucax: Other possible options:

* `Selected` -> `SelectedReceiver` (but it might be misleading because it is not really a receiver, just a wrapper)
* Remove `was_stopped` and force the user to use `exception` instead

Not sure why I'm listing them because I think I like them even less than the other options above :smile:

daniel-zullo-frequenz: >I was doing the change but then wondered what about exception and message, should they also be receiver_exception and receiver_message? 

Fair point. No, i don't think so. Considering those other properties then I'd say it's fine to keep the existent name `was_stopped` without prepending `receiver_`

llucax: @shsms ?

shsms: I'm not a fan of `was_stopped`, but that doesn't mean I can't accept it.

The issues I have with it are that it could mean other things - like the select was stopped/interrupted, etc.  Also the past tense `was` is also confusing.  Does that mean we don't know what the present state is?

And it doesn't even qualify that it should be checked within a type narrowed scope of `selected_from`.  It feels like it could be anything.

But of course, nothing wrong with keeping something ambiguous to discourage its use of force people to read the documentation.

shsms: The other fields don't need the `receiver_` prefix, because we don't have a `was_exception` or `was_message` to begin with.

llucax: I agree that `was_` is probably unnecessary, `stopped` is already a binary state, I think I used it because it read more nicely with `if selected.was_stopped` than `if selected.stopped`. Also I think the past tense is still correct. It was stopped when it was used to receive, we don't know if it got somehow restarted since then.

> And it doesn't even qualify that it should be checked within a type narrowed scope of `selected_from`. It feels like it could be anything.

I didn't get this.

> The other fields don't need the `receiver_` prefix, because we don't have a `was_exception` or `was_message` to begin with.

I don't get the difference. `exception` and `message` are actual objects, not binary states, so it doesn't make sense to have a `was_`. I thought you suggested the prefix to indicate that it's talking about the receiver, not the `Selected` object.

In any case, to wrap up, if you are not completely opposed to it and there is no consensus on a better choice, I'd say we just keep the status quo to be able to move forward.

llucax: I will merge this PR, as it is really a separate issue if we want to rename this or not. Feel free to create a issue for it @shsms and we can keep discussing it there.

PR #282, opened 2024-03-06
Improve generics usage

llucax:
This PR makes some improvements related to generic types.

It adds public generic type vars so the documentation is rendered more nicely and we get documentation on the type parameters.

It also extends the use of covariance for other classes or functions that work like containers that don't change the underlying data, so they can be used more flexibly.


llucax:
Based on, only the last 8 commits are relevant:
* #281

llucax:
Rebased to the current base branch.

src/frequenz/channels/_generic.py
```diff
@@ -0,0 +1,24 @@
...
+"""The type of the message received by the receiver after being mapped."""
+
+ReceiverMessageT_co = TypeVar("ReceiverMessageT_co", covariant=True)
+"""The type of the message received by a receiver."""
+
+SenderMessageT_co = TypeVar("SenderMessageT_co", covariant=True)
+"""The type of the message sent by a sender."""
+
+SenderMessageT_contra = TypeVar("SenderMessageT_contra", contravariant=True)
+"""The type of the message sent by a sender."""
```
daniel-zullo-frequenz, line 24: Should the docstring be a bit different here to highlight the subtle difference with  `SenderMessageT_co`? Otherwise the docstrings for `SenderMessageT_co` and `SenderMessageT_contra` is the same

RELEASE_NOTES.md
```diff
@@ -173,9 +173,11 @@
 
 ## Improvements
 
-* `Receiver`: Use a covariant generic type, which allows the generic type to be broader than the actual type.
+* `Receiver`, `merge`/`Merger`, `Error` and its derived classes now use a contravariant generic type, which allows the generic type to be narrower than the required type.
```
daniel-zullo-frequenz: I understood this should be a `covariant` generic type according to the code-changes. Can you please double-check?

daniel-zullo-frequenz (commented):
I have a couple of minor suggestions to check for. LGTM otherwise

src/frequenz/channels/_generic.py
llucax: I don't think so, this is an implementation detail the user doesn't have to care about. The user only needs to know that type is the type of message sent by the sender.

Also the suffixes `_co` and `_contra` are some sort of standard in Python, they are very well known (`pylint` even check for them) and the documentation still shows the full definition too:

image

RELEASE_NOTES.md
llucax: Yup, typo while shuffling stuff around, good catch!

llucax: Fixed.

shsms (dismissed):

RELEASE_NOTES.md
```diff
@@ -173,9 +173,11 @@
 
 ## Improvements
 
-* `Receiver`: Use a covariant generic type, which allows the generic type to be broader than the actual type.
+* `Receiver`, `merge`/`Merger`, `Error` and its derived classes now use a covariant generic type, which allows the generic type to be narrower than the required type.
```
daniel-zullo-frequenz: ?
```suggestion
* `Receiver`, `merge`/`Merger`, `Error` and its derived classes now use a covariant generic type, which allows the generic type to be broader than the required type.
```

llucax: Fixed :facepalm:

llucax:
Enabled auto-merge.

daniel-zullo-frequenz (approved):

PR #283, opened 2024-03-06
Improve `select()`-related exceptions

### LINKED ISSUE #280: Remove uses of `BaseException`
### What happened?

When we iterate over `select` without consuming a previously selected value, select raises a `SelectError` that derives from `BaseException`.

The use of `BaseException` combined with tasks consuming exceptions makes it hard to identify that anything happened, because a broad `except Exception:` block doesn't catch `BaseException`s.

The use of `BaseException` in user code is not recommended according to python docs: https://docs.python.org/3/library/exceptions.html#BaseException 

A `except Exception:` catches `select` errors.  i.e. `SelectError`s derive from `Exception` and not `BaseException`.

### Affected part(s)

Synchronization of multiple sources (`select()`, `merge()`, etc.) (part:synchronization)

llucax:
- **Don't inherit from `BaseException`**
- **Remove `SelectErrorGroup`**

Fixes #280.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -71,7 +71,6 @@
 
   - `Selected`
   - `SelectError`
-  - `SelectErrorGroup`
   - `UnhandledSelectedError`
   - `select`
   - `selected_from`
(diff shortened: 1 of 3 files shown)


llucax:
Enabled auto-merge.

shsms (approved):

PR #295, opened 2024-05-14
Fix the logger for the `Broadcast` channel

llucax:
The log messages were missing the level, timestamp, etc.) because we were using the wrong function to get the logger.


llucax:
Auto-merge enabled. I will go ahead and release v1.0.1 after this is merged.

shsms (approved):

PR #301, opened 2024-06-19
Add abstractions for sharing messages between channels

### LINKED ISSUE #300: Abstractions for sharing messages across channels
### What's needed?

There seem to be cases where it would be useful to link multiple channels - either sending the same message into multiple channels, or sending all messages received by a channel into another channel.

### Proposed solution

Implement:
- `TeeSender`, like the `tee` command.  Implements `Sender`, accepts multiple senders as constructor parameters, and sends messages sent to it to each of the destination senders.
- `Bridge` that takes a sender and a receiver, and send messages to the sender that it received from the receiver.

### Use cases

_No response_

_No response_

### Additional context

_No response_

shsms:
This PR implements two `experimental` features:

- `RelaySender`, which is a `Sender` that, forwards the messages sent to it, to multiple senders.
- `Pipe`, which provides a pipe between two channels, by connecting a `Receiver` to a `Sender`.

Closes #300

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,10 @@
 
 - The `LatestValueCache` class, which used to be internal to the Frequenz SDK, is now available through the channels package.
 
+- **Experimental**: `RelaySender`, which is a `Sender` that forwards the messages sent to it, to multiple senders.
+
+- **Experimental**: `Pipe`, which provides a pipe between two channels, by connecting a `Receiver` to a `Sender`.
+
 ## Bug Fixes
 
 <!-- Here goes notable bug fixes that are worth a special mention or explanation -->
--- a/src/frequenz/channels/experimental/__init__.py
+++ b/src/frequenz/channels/experimental/__init__.py
@@ -1,0 +1,12 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Experimental channel primitives.
+
+Warning:
+    This package contains experimental channel primitives that are not yet
+    considered stable. They are subject to change without notice, including
+    removal, even in minor updates.
+"""
+
+from ._pipe import Pipe
--- a/src/frequenz/channels/experimental/__init__.py
+++ b/src/frequenz/channels/experimental/__init__.py
@@ -10,3 +10,9 @@ Warning:
 """
 
 from ._pipe import Pipe
+from ._relay_sender import RelaySender
+
+__all__ = [
+    "Pipe",
+    "RelaySender",
+]
--- a/src/frequenz/channels/experimental/_pipe.py
+++ b/src/frequenz/channels/experimental/_pipe.py
@@ -1,0 +1,85 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Pipe between a receiver and a sender.
+
+The `Pipe` class takes a receiver and a sender and creates a pipe between them by
+forwarding all the messages received by the receiver to the sender.
+"""
+
+from __future__ import annotations
+
+import asyncio
+import typing
+
+from .._generic import ChannelMessageT
+from .._receiver import Receiver
+from .._sender import Sender
+
+
+class Pipe(typing.Generic[ChannelMessageT]):
+    """A pipe between two channels.
+
+    The `Pipe` class takes a receiver and a sender and creates a pipe between them
+    by forwarding all the messages received by the receiver to the sender.
+
+    Example:
+        ```python
+        from frequenz.channels import Broadcast, Pipe
+
+        channel1: Broadcast[int] = Broadcast(name="channel1")
+        channel2: Broadcast[int] = Broadcast(name="channel2")
+
+        receiver_chan1 = channel1.new_receiver()
+        sender_chan2 = channel2.new_sender()
+
+        async with Pipe(channel2.new_receiver(), channel1.new_sender()):
+            await sender_chan2.send(10)
+            assert await receiver_chan1.receive() == 10
+        ```
+    """
+
+    def __init__(
+        self, receiver: Receiver[ChannelMessageT], sender: Sender[ChannelMessageT]
+    ) -> None:
+        """Create a new pipe between two channels.
+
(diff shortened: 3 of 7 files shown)


llucax:
I made some comments in the issue.

shsms:
> I made some comments in the issue.

Replied in the issue.

src/frequenz/channels/experimental/__init__.py
```diff
@@ -0,0 +1,16 @@
+# License: MIT
+# Copyright © 2024 Frequenz Energy-as-a-Service GmbH
+
+"""Experimental channel primitives.
+
+This package contains experimental channel primitives that are not yet
+considered stable. They are subject to change without notice.
```
llucax: Maybe like this to make it more scary?

```suggestion
Warning:
    This package contains experimental channel primitives that are not yet
    considered stable. They are subject to change or removal without
    notice even in minor updates.
```

tests/experimental/test_pipe.py
```diff
@@ -0,0 +1,57 @@
...
+        assert await receive_timeout(receiver_chan1) == 42
+        assert await receive_timeout(receiver_chan2) == 42
+
+        await sender_chan2.send(-2)
+        assert await receive_timeout(receiver_chan1) == -2
+        assert await receive_timeout(receiver_chan2) == -2
+
+        await sender_chan1.send(43)
+        assert await receive_timeout(receiver_chan1) == 43
+        assert await receive_timeout(receiver_chan2) is Timeout
```
llucax, line 49: I guess you already know, but just in case, another way to do this is:

```py
async def receive_timeout(recv: Receiver[T], timeout: float = 0.1) -> T | type[Timeout]:
    """Receive message from receiver with timeout."""
    return await asyncio.wait_for(recv.receive(), timeout=timeout)
        
with pytest.raises(TimeoutError):
    await receive_timeout(receiver_chan2)
```

I guess you chose this approach because it is slightly more compact?

RELEASE_NOTES.md
```diff
@@ -12,6 +12,10 @@
 
 - The `LatestValueCache` class, which used to be internal to the Frequenz SDK, is now available through the channels package.
 
+- `RelaySender`, which is a `Sender` that forwards the messages sent to it, to multiple senders.
+
+- `Pipe`, which provides a pipe between two channels, by connecting a `Receiver` to a `Sender`.
```
llucax: Maybe `- **Experimental** ...`?

src/frequenz/channels/experimental/__init__.py
```diff
@@ -0,0 +1,16 @@
...
+"""
+
+from ._pipe import Pipe
+from ._relay_sender import RelaySender
+
+__all__ = [
+    "Pipe",
```
llucax, line 16: Nitpick: This line was added in the `RelaySender` commit instead of the `Pipe` commit.

src/frequenz/channels/experimental/_pipe.py
```diff
@@ -0,0 +1,85 @@
...
+    Example:
+        ```python
+        from frequenz.channels import Broadcast, Pipe
+
+        channel1: Broadcast[int] = Broadcast(name="channel1")
+        channel2: Broadcast[int] = Broadcast(name="channel2")
+
+        async with Pipe(channel2.new_receiver(), channel1.new_sender()):
+            sender_chan2 = channel2.new_sender()
+            receiver_chan1 = channel1.new_receiver()
```
llucax: This could be also created outside the `async with`, right? It got me a bit confused because somehow I was expecting that `Pipe` was an object that you use more actively, like `pipe.send()` and `pipe.receive()`, like the `RelaySender`. I guess it doesn't make sense to implement it like that, but that's what I had in mind while talking about it for some reason :facepalm: 

I would move it outside to make it more clear.

----

Probably not for this PR, more like brainstorming:

With this in mind, and if this is mostly intended to be used as a context manager, then maybe we could create a function called `connecting()` that creates the `Pipe` (name inspired by [closing()](https://docs.python.org/3/library/contextlib.html#contextlib.closing)) context manger.

```py
async with connecting(channel2.new_receiver(), channel1.new_sender()):
    await sender_chan2.send(42)
    assert await receiver_chan1.receive() == 42
```

(it could be `piping()` too but IMHO it sounds weirder as a verb, and `connecting` seems clear enough).

Also, I understand this is useful as it is, and it does what it says it does, but for me it is a bit confusing that is like having a side effect, you are not using the sender and receiver that you make part of the `Pipe`, is more like connecting channels (I mean the end effect), and this depends on the type of channel. I don't have a proposal for solving this, just mentioning that got me confused for a bit. I thought of an options that makes it more clear to me but I guess it's not suitable: 

Having a `async with connecting(source: Broadcast, destination: Broadcast):`, this for me sounds much more clear, but I guess we probably want to use this in a place where we just have a sender and a receiver, not the channels.

llucax (dismissed):
Only a few very minor and optional comments, and one very wide brainstorming topic that I don't think it should be addressed in this PR :P, so approving.

src/frequenz/channels/experimental/__init__.py
shsms: Done with slightly different wording.

tests/experimental/test_pipe.py
shsms: No, the compactness was just a side-effect.  It is because we're not testing if there was an exception, we don't care about that, we just want to test that no messages were received.

RELEASE_NOTES.md
shsms: done

src/frequenz/channels/experimental/__init__.py
shsms: yes, I was very self-conscious when I was doing it, but I guess my laziness was stronger.

src/frequenz/channels/experimental/_pipe.py
shsms: > I would move it outside to make it more clear.

done.

llucax (commented):
No new review requested yet, but re-approving anyways.

llucax (approved):

PR #303, opened 2024-06-27
Support filtering the messages on a receiver

shsms:
A new `filter` method is added to the `Receiver` interface, which allows the application of a filter function on the messages on a receiver.

Example:

```python
async for message in receiver.filter(lambda num: num % 2):
    print(f"An even number: {message}")
```

### DIFF
--- a/src/frequenz/channels/_receiver.py
+++ b/src/frequenz/channels/_receiver.py
@@ -242,6 +242,25 @@ class Receiver(ABC, Generic[ReceiverMessageT_co]):
         """
         return _Mapper(receiver=self, mapping_function=mapping_function)
 
+    def filter(
+        self, filter_function: Callable[[ReceiverMessageT_co], bool], /
+    ) -> Receiver[ReceiverMessageT_co]:
+        """Apply a filter function on the messages on a receiver.
+
+        Tip:
+            The returned receiver type won't have all the methods of the original
+            receiver. If you need to access methods of the original receiver that are
+            not part of the `Receiver` interface you should save a reference to the
+            original receiver and use that instead.
+
+        Args:
+            filter_function: The function to be applied on incoming messages.
+
+        Returns:
+            A new receiver that applies the function on the received messages.
+        """
+        return _Filter(receiver=self, filter_function=filter_function)
+
 
 class ReceiverError(Error, Generic[ReceiverMessageT_co]):
     """An error that originated in a [Receiver][frequenz.channels.Receiver].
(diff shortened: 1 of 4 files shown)


llucax:
What about adding an `async_filter` to the new core repo instead? This would be useful with any async iterator. I would also remove `map()` and add an `async_map()` to core.

llucax:
There is also https://snyk.io/advisor/python/asyncstdlib. Maybe we should start using that? It is strange that this is not part of Python stdlib already.

llucax:
https://asyncstdlib.readthedocs.io/en/latest/source/api/builtins.html#asyncstdlib.builtins.filter

llucax:
If we agree it is a good idea we could add it to repo-config so it is used by default, like we do with `typing_extensions`.

shsms:
> What about adding an async_filter to the new core repo instead? This would be useful with any async iterator. I would also remove map() and add an async_map() to core.

This gives out `Receiver`s, we can do all the things we can do with receivers with them.  Maybe we can have an additional `async_filter` or use `asyncstd`, but having it here is still valuable.

shsms:
> If we agree it is a good idea we could add it to repo-config so it is used by default, like we do with typing_extensions.

I do agree, it would be very useful, but I don't think it can replace this PR, because the filter in this PR returns receivers, which we can put in select or any of the many possibilities.

llucax:
Right. OK, and I guess it is useful to get a receiver because it can be used in `select()`. It came to my mind a couple of times that maybe we should have a utility receiver that converts any async iterator into a receiver. So we could potentially do:

```py
filtered_receiver = as_receiver(a.filter(lambda num: num % 2), original_receiver)
async for selected in select(filtered_receiver, ...):
    print(f"An even number: {message}")
```

But if we want to move in that direction, we'll eventually have to remove `map()` too, and that will be a breaking change, so I'm happy to leave that for 2.0 and add `filter` now if it is useful.

Will start the review now then...

llucax:
> > What about adding an async_filter to the new core repo instead? This would be useful with any async iterator. I would also remove map() and add an async_map() to core.
> 
> This gives out `Receiver`s, we can do all the things we can do with receivers with them. Maybe we can have an additional `async_filter` or use `asyncstd`, but having it here is still valuable.

* https://github.com/frequenz-floss/frequenz-repo-config-python/issues/282

src/frequenz/channels/_receiver.py
```diff
@@ -342,3 +361,79 @@ def __str__(self) -> str:
...
+        """
+        if self._recv_closed:
+            raise ReceiverStoppedError(self)
+        assert not isinstance(
+            self._next_message, _Sentinel
+        ), "`consume()` must be preceded by a call to `ready()`"
+
+        message = self._next_message
+        self._next_message = _SENTINEL
+        return message
```
llucax, line 443: It would be good also to add a `__repr__` and `__str__` here too, including the original receiver so one can know the source of this receiver. You can probably copy them from `_Mapper`.

src/frequenz/channels/_receiver.py
```diff
@@ -342,3 +361,79 @@ def __str__(self) -> str:
...
         """Return a string representation of the timer."""
         return f"{type(self).__name__}({self._receiver!r}, {self._mapping_function!r})"
+
+
+class _Sentinel:
+    """A sentinel object to represent no value received yet."""
+
+    def __repr__(self) -> str:
+        """Return a string representation of this sentinel."""
+        return "Sentinel()"
```
llucax: Maybe it is worth adding a `__str__` too in case this is printed somewhere like in `LastValueCache`? I when adding the `__str__` and `__repr__` for `_Filter` it will be used.

src/frequenz/channels/_receiver.py
shsms: done.

src/frequenz/channels/_receiver.py
shsms: Added a `__str__`, but also removed `__repr__` for the sentinel.

shsms:
Based on #301

src/frequenz/channels/_receiver.py
```diff
@@ -336,9 +355,93 @@ def consume(self) -> MappedMessageT_co:  # noqa: DOC502
...
+            self._next_message, _Sentinel
+        ), "`consume()` must be preceded by a call to `ready()`"
+
+        message = self._next_message
+        self._next_message = _SENTINEL
+        return message
+
+    def __str__(self) -> str:
+        """Return a string representation of the filter."""
+        return f"{type(self).__name__}:{self._receiver}:{self._filter_function}"
```
llucax, line 447: Wouldn't it be useful here to print also the `_next_message`?

shsms: I think it doesn't make sense, because it would only have a value between calls to `ready` and `consume`.  At all other times, it would only print the sentinel.  So I don't think there's any value in it.

llucax: Right, is not like `LatestValueCache`.

llucax: But then if the `Sentinel` class is never printed nor exposed outside, there is not much point in having either of `__str__` nor `__repr__`, unless we use it for internal debugging, in which case I think it makes more sense to have `__repr__` as normally pytest and debuggers will use that one.

I think I would print the `_next_value` in this classes `__repr__`, as when debugging or in a pytest it might be useful to see what are its contents, even if most of the time is the sentinel. When calling `repr()` you usually want to know the internal state of the instance. Usually `<` `>` is used to represent internal state when `repr()` doesn't match the code used to create the class, like the default `__repr__` in object and what's used to print functions or methods:

```py
>>> class C: pass
... 
>>> repr(C())
'<__main__.C object at 0x7f564f233690>'
>>> repr(C.__init__)
"<slot wrapper '__init__' of 'object' objects>"
```

shsms: done

llucax (commented):
LGTM. Not approving until it's not a draft anymore.

llucax:
Also tests are failing:
```
src/frequenz/channels/_receiver.py:451:101: E501 line too long (138 > 100 characters)
```

shsms:
Fixed the long line

llucax (approved):

PR #318, opened 2024-09-09
Fix ready() method in FileWatcher

### LINKED ISSUE #317: FileWatcher stops working after exiting async iterator
### What happened?

The `FileWatcher` breaks after exiting an async iterator when the task running it gets cancelled, `FileWatcher` enters an infinite loop where `ready()` continuously returns False, and `consume()` raises an error. The underlying issue seems to be that `self._awatch_stopped_exc` is never cleared after cancellation and `self._awatch` seems broken.

The `FileWatcher` should continue working if it is reused in a new async iterator, without needing to be re-instantiated.

### Affected version(s)

_No response_

### Affected part(s)

Utilities (`Event`, `FileWatcher`, `Timer`, etc.) (part:utilities)

### Extra information

The following script reproduces the issue and is similar to what the `ConfigManagingActor` in the frequenz-SDK does. If the actor is restarted for any reason, the `FileWatcher` stops working and only works again if re-created.

```python
import asyncio
from datetime import timedelta
from pathlib import Path

from frequenz.channels import select, selected_from
from frequenz.channels.file_watcher import EventType, FileWatcher
from frequenz.channels.timer import SkipMissedAndDrift, Timer


async def read(file_path, file_watcher) -> None:
    """Read file events."""
    timer_write = Timer(timedelta(seconds=1), SkipMissedAndDrift())

    async for selected in select(file_watcher, timer_write):
        if selected_from(selected, file_watcher):
            print(f"New file event: {selected.message}")
        elif selected_from(selected, timer_write):
            file_path.write_text(f"{selected.message}")
            print(f"Writing {selected.message} to file: {file_path}")
        else:
            print("Unknown selected")


async def run():
    file_path = Path("config.txt")

    file_watcher = FileWatcher(
        paths=[file_path.parent], event_types=frozenset(EventType)
    )

    task = asyncio.create_task(read(file_path, file_watcher))

    timer_stop = Timer(timedelta(seconds=3), SkipMissedAndDrift())

    while True:
        await timer_stop.receive()
        task.cancel()
        print("Task cancelled. New task will be created.")
        await asyncio.sleep(0.1)
        task = asyncio.create_task(read(file_path, file_watcher))


asyncio.run(run())
```

daniel-zullo-frequenz:
The ready() method now returns False when an exception is set.

Fixes #317

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-- `FileWatcher`: Fixed `ready()` method to return False when an error occurs.
+- `FileWatcher`: Fixed `ready()` method to return False when an error occurs. Before this fix, `select()` (and other code using `ready()`) never detected the `FileWatcher` was stopped and the `select()` loop was continuously waking up to inform the receiver was ready.
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- `FileWatcher`: Fixed `ready()` method to return False when an error occurs.
--- a/src/frequenz/channels/file_watcher.py
+++ b/src/frequenz/channels/file_watcher.py
@@ -185,6 +185,7 @@ class FileWatcher(Receiver[Event]):
             self._changes = await anext(self._awatch)
         except StopAsyncIteration as err:
             self._awatch_stopped_exc = err
+            return False
 
         return True
 
--- a/src/frequenz/channels/file_watcher.py
+++ b/src/frequenz/channels/file_watcher.py
@@ -70,6 +70,15 @@ class FileWatcher(Receiver[Event]):
     the [`path`][frequenz.channels.file_watcher.Event.path] where the change was
     observed.
 
+    Note:
+        The owner of the [`FileWatcher`][frequenz.channels.file_watcher.FileWatcher]
+        receiver is responsible for recreating the `FileWatcher` after it has been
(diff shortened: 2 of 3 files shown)


llucax (commented):
The diff LGTM, but the PR still needs some improvements:

* Update the PR title as this only fix a bug now
* Update the PR body as it will be part of the merge commit, it doesn't make sense to have outdated, struck out, information.
* It needs release notes

daniel-zullo-frequenz:
Updated PR tittle and description. The release-notes was already added, do you want me to change something there?

llucax:
Woops, I guess I was looking commit-by-commit and forget to hit next at some point.

RELEASE_NOTES.md
```diff
@@ -14,4 +14,4 @@
 
 ## Bug Fixes
 
-<!-- Here goes notable bug fixes that are worth a special mention or explanation -->
+- `FileWatcher`: Fixed `ready()` method to return False when an error occurs.
```
```suggestion
- `FileWatcher`: Fixed `ready()` method to return False when an error occurs. Before this fix, `select()` (and other code using `ready()`) never detected the `FileWatcher` was stopped and the `select()` loop was continuously waking up to inform the receiver was ready.
```

llucax (dismissed):
Just an optional improvement over the release notes, trying to explain a bit more the implications of the bug, which is what users care more about (just knowing that a `return False` was missed in a method they almost never use is not that informative).

daniel-zullo-frequenz:
> Just an optional improvement over the release notes, trying to explain a bit more the implications of the bug, which is what users care more about (just knowing that a return False was missed in a method they almost never use is not that informative).

Thanks! Added to suggestion to  the release-notes

llucax (approved):

PR #320, opened 2024-09-17
Add scheduling class optimized for frequently changing timers

Marenz:
80% ChatGPT-o1-preview, 20% me

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,5 +1,9 @@
 # Frequenz channels Release Notes
 
+## New Features
+
+- A new class `frequenz.channels.time_scheduler.TimeScheduler` was added. It's optimized for scenarios where events get added, rescheduled or canceled frequently.
+
 ## Bug Fixes
 
 - `FileWatcher`: Fixed `ready()` method to return False when an error occurs. Before this fix, `select()` (and other code using `ready()`) never detected the `FileWatcher` was stopped and the `select()` loop was continuously waking up to inform the receiver was ready.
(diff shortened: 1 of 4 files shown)


llucax:
I'm not convinced this belongs in this repo, and I'm not convinced by the interface either. Bad ChatGPT, bad!

Seriously, I would much rather make `interval` changable via `reset()` in the current timer, and build something on top of that instead of re-implementing the whole thing. All we need is a interval-reseteable timer and a container that gives you the next interval from a list of "scheduled timers", they can be completely decoupled.

If there is no such a container available yet, I would add it to https://github.com/frequenz-floss/frequenz-core-python/ instead.

What I envision is something like:

```py
timer_scheduler = TimerScheduler()
timer_scheduler.add(...)

timer = Timer(timer_scheduler.next(), ...)

async for selected in select(timer, other):
    if selected_from(selected, timer):
        ...
        timer.reset(timer_scheduler.next())
    if selected_from(selected, other):
        ...
        timer_scheduler.add(...) # or timer_scheduler.remove(...)    
...
```

Marenz:
> I'm not convinced this belongs in this repo, and I'm not convinced by the interface either. Bad ChatGPT, bad!

Well, I designed the interface and ChatGPT did a good job at it imo.
And I like it too.

> What I envision is something like:

How would you propose to interrupt a scheduled timer that is currently in `ready()`'s `await asyncio.sleep(..)`?
Which task do we have to cancel?

llucax:
> How would you propose to interrupt a scheduled timer that is currently in `ready()`'s `await asyncio.sleep(..)`?
> Which task do we have to cancel?

I guess you already answer it for yourself with #320 :)

PR #321, opened 2024-09-18
Timer: Add support for rearming timer with new interval

Marenz:
Ugly alternative to #320

>They eat our timers, our start_delays!

### DIFF
--- a/src/frequenz/channels/timer.py
+++ b/src/frequenz/channels/timer.py
@@ -586,7 +586,12 @@ class Timer(Receiver[timedelta]):
         """Whether the timer is running."""
         return not self._stopped
 
-    def reset(self, *, start_delay: timedelta = timedelta(0)) -> None:
+    def reset(
+        self,
+        *,
+        interval: timedelta | None = None,
+        start_delay: timedelta = timedelta(0),
+    ) -> None:
         """Reset the timer to start timing from now (plus an optional delay).
 
         If the timer was stopped, or not started yet, it will be started.
@@ -595,6 +600,8 @@ class Timer(Receiver[timedelta]):
         more details.
 
         Args:
+            interval: The new interval between ticks. If `None`, the current
+                interval is kept.
             start_delay: The delay before the timer should start. This has microseconds
                 resolution, anything smaller than a microsecond means no delay.
 
@@ -606,6 +613,9 @@ class Timer(Receiver[timedelta]):
 
         if start_delay_ms < 0:
             raise ValueError(f"`start_delay` can't be negative, got {start_delay}")
+
+        if interval is not None:
+            self._interval = _to_microseconds(interval)
 
         self._next_tick_time = self._now() + start_delay_ms + self._interval
 
--- a/src/frequenz/channels/timer.py
+++ b/src/frequenz/channels/timer.py
@@ -523,6 +523,8 @@ class Timer(Receiver[timedelta]):
         See the documentation of `MissedTickPolicy` for details.
         """
 
+        self._reset_event = asyncio.Event()
+
         self._loop: asyncio.AbstractEventLoop = (
             loop if loop is not None else asyncio.get_running_loop()
         )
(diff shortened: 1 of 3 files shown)


src/frequenz/channels/timer.py
```diff
@@ -605,7 +614,16 @@ def reset(self, *, start_delay: timedelta = timedelta(0)) -> None:
...
             raise ValueError(f"`start_delay` can't be negative, got {start_delay}")
         self._stopped = False
+
+        if interval is not None:
+            self._interval = _to_microseconds(interval)
+
+        prev_next_tick_time = self._next_tick_time
         self._next_tick_time = self._now() + start_delay_ms + self._interval
+
+        if prev_next_tick_time and prev_next_tick_time > self._next_tick_time:
```
llucax: This check is to know if this is the first time the timer is started? I think it is worth a comment. Also for the `>` condition, as it wasn't obvious for me (I hope I'm getting it right).

```suggestion
        # If there was no previous `next_tick_time`, it means the timer wasn't started yet, so no need to
        # fire the reset event. If the new `_next_tick_time` is bigger than the previous one, we also don't
        # need to fire the reset even, as `ready()` will keep looping until the new `_next_tick_time` is
        # reached.
        if prev_next_tick_time and prev_next_tick_time > self._next_tick_time:
```

src/frequenz/channels/timer.py
```diff
@@ -664,7 +683,12 @@ async def ready(self) -> bool:  # noqa: DOC502
         # could be reset while we are sleeping, in which case we need to recalculate
         # the time to the next tick and try again.
         while time_to_next_tick > 0:
-            await asyncio.sleep(time_to_next_tick / 1_000_000)
+            for fin in asyncio.as_completed(
```
llucax: What does `fin` stands for? Why not `completed` or something more obvious?

src/frequenz/channels/timer.py
```diff
@@ -664,7 +683,12 @@ async def ready(self) -> bool:  # noqa: DOC502
...
+                [asyncio.sleep(time_to_next_tick / 1_000_000), self._reset_event.wait()]
+            ):
+                await fin
+                break
```
llucax: I was a bit confused by this `break` too (I somehow thought it was `break`ing the outer `while` instead of the `for`).

Maybe it could be a bit more clear like this?
```suggestion
            await next(asyncio.as_completed(
                [asyncio.sleep(time_to_next_tick / 1_000_000), self._reset_event.wait()]
            ))
```

llucax (commented):
Just a couple of minor comments to improve clarity, as it wasn't trivial for me to understand all that was going on.

Also I think this bug fix and new feature deserve some tests.

src/frequenz/channels/timer.py
Marenz: Funny enough, I didn't even think about it, I was just making mypy happy :laughing:

Marenz: I was mainly checking if the new time is ealier or not like you described.
But I think we can simplify this.

src/frequenz/channels/timer.py
```diff
@@ -604,8 +613,16 @@ def reset(self, *, start_delay: timedelta = timedelta(0)) -> None:
...
             raise ValueError(f"`start_delay` can't be negative, got {start_delay}")
-        self._stopped = False
+
+        if interval is not None:
+            self._interval = _to_microseconds(interval)
+
         self._next_tick_time = self._now() + start_delay_ms + self._interval
+
+        if self.is_running:
+            self._reset_event.set()
```
llucax, line 623: Just out of curiosity, was the `if` necessary or is it just an *optimization*? Going through it quickly it looks like it should work without it too (if the timer is not running, it will just break the wait loop but it should resume normally, afterwards, right?), but maybe I'm missing some edge case.

llucax (dismissed):

llucax:
Not sure how strict we should be about this, but the bug fix should be applied to v1.1.x, but the new feature should go to v1.2.0.

I think to have it both and not being **that** bureaucratic, you could split it in 2 PRs, bug fix and feature, and we merge first the bug fix, release v1.1.3, then merge the feature and release v1.2.0.

llucax:
For simplicity I think we can release all of this in v1.2.0 and that's it.

llucax (approved):

src/frequenz/channels/timer.py
Marenz: We would probably be fine if we just set it every time, yes.

PR #322, opened 2024-09-18
Enable polling in file watcher

### LINKED ISSUE #256: Filewatcher is not reporting events in CIFS mounts
### What happened?

A lot of cloud use-cases for FileWatcher need it to be monitoring NFS/CIFS mounts, which don't emit inotify events on remote updates.

So our FileWatcher which uses the `watchfiles` library, doesn't emit events either.  An alternative is for the FileWatcher to use the [watchdog](https://github.com/gorakhargosh/watchdog#about-using-watchdog-with-cifs) library, which also has a polling observer.

FS events triggered remotely also produce a message from the FileWatcher.

### Affected version(s)

_No response_

### Affected part(s)

Utility receivers, `Event`, `FileWatcher`, `Timer`, etc. (part:receivers)

### Extra information

_No response_

daniel-zullo-frequenz:
`inotify` does not work reliably with network file systems (e.g., NFS, CIFS) commonly used in cloud environments. These
file systems may not propagate file system events correctly, causing `inotify` to miss changes. To ensure consistent file
monitoring across these environments, polling is enabled by default in `FileWatcher`.

Fixes #256

### DIFF
--- a/.github/containers/test-installation/Dockerfile
+++ b/.github/containers/test-installation/Dockerfile
@@ -3,7 +3,7 @@
 # This Dockerfile is used to test the installation of the python package in
 # multiple platforms in the CI. It is not used to build the package itself.
 
-FROM --platform=${TARGETPLATFORM} python:3.11-slim
+FROM python:3.11-slim
 
 RUN apt-get update -y && \
     apt-get install --no-install-recommends -y \
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -1,11 +1,19 @@
 # Frequenz channels Release Notes
 
+## Upgrading
+
+- `FileWatcher`: The file polling mechanism is now forced by default. This provides reliable and consistent file monitoring on network file systems (e.g., CIFS). However, it may have a performance impact on local file systems or when monitoring a large number of files.
+  - To disable file polling, set the `force_polling` parameter to `False`.
+  - The `polling_interval` parameter defines the interval for polling changes. This is relevant only when polling is enabled and defaults to 1 second.
+
 ## New Features
 
 - `Timer.reset()` now supports setting the interval and will restart the timer with the new interval.
 
 ## Bug Fixes
 
-- `FileWatcher`: Fixed `ready()` method to return False when an error occurs. Before this fix, `select()` (and other code using `ready()`) never detected the `FileWatcher` was stopped and the `select()` loop was continuously waking up to inform the receiver was ready.
+- `FileWatcher`:
+  - Fixed `ready()` method to return False when an error occurs. Before this fix, `select()` (and other code using `ready()`) never detected the `FileWatcher` was stopped and the `select()` loop was continuously waking up to inform the receiver was ready.
+  - Reports file events correctly on network file systems like CIFS.
 
 - `Timer.stop()` and `Timer.reset()` now immediately stop the timer if it is running. Before this fix, the timer would continue to run until the next interval.
--- a/src/frequenz/channels/file_watcher.py
+++ b/src/frequenz/channels/file_watcher.py
@@ -22,6 +22,7 @@ import asyncio
 import pathlib
 from collections import abc
 from dataclasses import dataclass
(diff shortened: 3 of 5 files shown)


src/frequenz/channels/file_watcher.py
```diff
@@ -127,13 +128,19 @@ def __init__(
         self,
         paths: list[pathlib.Path | str],
         event_types: abc.Iterable[EventType] = frozenset(EventType),
+        *,
+        enable_polling: bool = True,
```
llucax: If you are using a positive term then I would use the same as `watchfiles` (`force_polling`), as probably even when passing `force_polling=False` (`enable_polling=False`) polling will probably be used if I notify or other event based system calls are not available in the OS.

tests/test_file_watcher_integration.py
llucax, line 1: Why disabling polling in tests? Are there any tests using polling?

daniel-zullo-frequenz: The unit test is using polling. I can try to make one integration tests work at least after introducing the `polling_interval` parameter.

daniel-zullo-frequenz:
Updated based on initial suggestions from Luca

daniel-zullo-frequenz:
Added a new commit to remove the redundant `TARGETPLATFORM` in Dockerfile

.github/containers/test-installation/Dockerfile
```diff
@@ -3,7 +3,7 @@
 # This Dockerfile is used to test the installation of the python package in
 # multiple platforms in the CI. It is not used to build the package itself.
 
-FROM --platform=${TARGETPLATFORM} python:3.11-slim
+FROM python:3.11-slim
```
llucax, line 6: :heart:

llucax (commented):
I have the same concern as with #321 , this is actually mixing one *bug fix* (setting `force_poll=True`, which depending on how we see it it could even be a breaking change[^1]) and a new feature (being able to configure `force_polling`).

[^1]: Maybe this is worth documenting in the `Upgrading` section, mentioning the default mechanism was changed and updates might have a bit of a delay now and watching a large amount of files might be less efficient, and letting the user know how to go back to the previous default.

llucax:
For simplicity I think we can release all of this in v1.2.0 and that's it.

llucax (dismissed):

daniel-zullo-frequenz:
Updated release-notes to add an entry in `Upgrading`. And rebased onto the latest v1.x.x

RELEASE_NOTES.md
```diff
@@ -1,11 +1,19 @@
 # Frequenz channels Release Notes
 
+## Upgrading
+
+- `FileWatcher`: The file polling mechanism is now enabled by default. This provides reliable and consistent file monitoring on network file systems (e.g., CIFS). However, it may have a performance impact on local file systems or when monitoring a large number of files.
```
```suggestion
- `FileWatcher`: The file polling mechanism is now forced by default. This provides reliable and consistent file monitoring on network file systems (e.g., CIFS). However, it may have a performance impact on local file systems or when monitoring a large number of files.
```

tests/test_file_watcher_integration.py
```diff
@@ -113,7 +117,11 @@ async def test_file_watcher_exit_iterator(tmp_path: pathlib.Path) -> None:
     number_of_writes = 0
     expected_number_of_writes = 3
 
-    file_watcher = FileWatcher(paths=[str(tmp_path)])
+    file_watcher = FileWatcher(
+        paths=[str(tmp_path)],
+        force_polling=True,
+        polling_interval=timedelta(seconds=0.05),
```
llucax, line 123: 50ms sounds a bit extreme, but I guess it's fine...

daniel-zullo-frequenz: I know, the timer of the test is set to 100ms  so it needed a safe value to ensure the changes are picked. We can increase the timer interval but that'd make the test to take longer

llucax (approved):

PR #323, opened 2024-09-23
Clean up `Timer` tasks

Marenz:


llucax (commented):
I don't think this is necessary, if you are seeing warning in `pytest` about tasks being pending, is a bug in pytest: https://github.com/pytest-dev/pytest-asyncio/issues/200

Unless I'm missing something, I think `as_completed()` is being used properly here.

Marenz:
It might not be necessary, but I believe it actually speeds up runtime if we clean up our tasks afterwards.

llucax:
Why do you think that? But in any case, if you want to do eager finalization (so not letting the GC pick up your garbage), then probably this is shorter and should do the same:

```py
            completed = asyncio.as_completed(
                [asyncio.sleep(time_to_next_tick / 1_000_000), self._reset_event.wait(),]
            )

            try:
                await next(completed)
            finally:
                del completed
```

Marenz:
Well, now I did the actual measurement and it does not seem to help at all XD

PR #326, opened 2024-09-25
Fix file watcher integration tests

### LINKED ISSUE #324: `tests/test_file_watcher_integration.py` became flaky after using `force_polling=True`
### What happened?

We are having some sporadic failures, specifically in **cross-arch tests** (both `pytest_min` and `pytest_max`), for file watcher integration tests. In particular this test seems to be flaky: [`tests/test_file_watcher_integration.py::test_file_watcher`](https://github.com/frequenz-floss/frequenz-channels-python/blob/13cf1741cb4234ef529e46c593c390f2730111bc/tests/test_file_watcher_integration.py#L18-L43).

```
=================================== FAILURES ===================================
______________________________ test_file_watcher _______________________________

tmp_path = PosixPath('/tmp/pytest-of-root/pytest-0/test_file_watcher0')

    @pytest.mark.integration
    async def test_file_watcher(tmp_path: pathlib.Path) -> None:
        """Ensure file watcher is returning paths on file events.
    
        Args:
            tmp_path: A tmp directory to run the file watcher on. Created by pytest.
        """
        filename = tmp_path / "test-file"
    
        number_of_writes = 0
        expected_number_of_writes = 3
    
        file_watcher = FileWatcher(paths=[str(tmp_path)])
        timer = Timer(timedelta(seconds=0.1), SkipMissedAndDrift())
    
        async for selected in select(file_watcher, timer):
            if selected_from(selected, timer):
                filename.write_text(f"{selected.message}")
            elif selected_from(selected, file_watcher):
                event_type = EventType.CREATE if number_of_writes == 0 else EventType.MODIFY
>               assert selected.message == Event(type=event_type, path=filename)
E               AssertionError: assert Event(type=<EventType.MODIFY: <Change.modified: 2>>, path=PosixPath('/tmp/pytest-of-root/pytest-0/test_file_watcher0')) == Event(type=<EventType.CREATE: <Change.added: 1>>, path=PosixPath('/tmp/pytest-of-root/pytest-0/test_file_watcher0/test-file'))
E                 
E                 Differing attributes:
E                 ['type', 'path']
E                 
E                 Drill down into differing attribute type:
E                   type: <EventType.MODIFY: <Change.modified: 2>> != <EventType.CREATE: <Change.added: 1>>
E                 
E                 Drill down into differing attribute path:
E                   path: PosixPath('/tmp/pytest-of-root/pytest-0/test_file_watcher0') != PosixPath('/tmp/pytest-of-root/pytest-0/test_file_watcher0/test-file')

tests/test_file_watcher_integration.py:37: AssertionError
```

Test to always pass.

### Affected version(s)

v1.2.0

### Affected part(s)

Unit, integration and performance tests (part:tests)

### Extra information

* Failed jobs

    * https://github.com/frequenz-floss/frequenz-channels-python/actions/runs/11030029977/job/30633740881
    * https://github.com/frequenz-floss/frequenz-channels-python/actions/runs/10995292713/attempts/1

* Tips

    * We are using some small times in those tests, in the order of [tens](https://github.com/frequenz-floss/frequenz-channels-python/blob/13cf1741cb4234ef529e46c593c390f2730111bc/tests/test_file_watcher_integration.py#L123) or [hundreds](https://github.com/frequenz-floss/frequenz-channels-python/blob/13cf1741cb4234ef529e46c593c390f2730111bc/tests/test_file_watcher_integration.py#L30) of milliseconds, since cross-arch tests are **sloooooow** the flakiness probably comes from there.

llucax:
When watching the whole test directory we'll also get some events for the directory, and those events can come with different order depending on the OS, so we can't rely on them. We check that if a directory event comes, it is a modification.

Also when testing for events in a file watcher, we need to set the `polling_interval` to a value that is small enough to detect the changes we are expecting, otherwise some events could be "merged" if the changes happen faster than the polling interval.

Fixes #324.

### DIFF
--- a/tests/test_file_watcher_integration.py
+++ b/tests/test_file_watcher_integration.py
@@ -26,7 +26,9 @@ async def test_file_watcher(tmp_path: pathlib.Path) -> None:
     number_of_writes = 0
     expected_number_of_writes = 3
 
-    file_watcher = FileWatcher(paths=[str(tmp_path)])
+    file_watcher = FileWatcher(
+        paths=[str(tmp_path)], polling_interval=timedelta(seconds=0.05)
+    )
     timer = Timer(timedelta(seconds=0.1), SkipMissedAndDrift())
 
     async for selected in select(file_watcher, timer):
@@ -65,6 +67,7 @@ async def test_file_watcher_deletes(tmp_path: pathlib.Path) -> None:
         paths=[str(tmp_path)],
         event_types={EventType.DELETE},
         force_polling=False,
+        polling_interval=timedelta(seconds=0.05),
     )
     write_timer = Timer(timedelta(seconds=0.1), SkipMissedAndDrift())
     deletion_timer = Timer(timedelta(seconds=0.25), SkipMissedAndDrift())
--- a/tests/test_file_watcher_integration.py
+++ b/tests/test_file_watcher_integration.py
@@ -10,7 +10,7 @@ from datetime import timedelta
 import pytest
 
 from frequenz.channels import ReceiverStoppedError, select, selected_from
-from frequenz.channels.file_watcher import Event, EventType, FileWatcher
+from frequenz.channels.file_watcher import EventType, FileWatcher
 from frequenz.channels.timer import SkipMissedAndDrift, Timer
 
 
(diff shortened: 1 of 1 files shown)


llucax:
Setting auto-merge.

Marenz (approved):

PR #328, opened 2024-10-08
Add a `Receiver.triggered` method

shsms:
This is an alternative to the `selected_from` method.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -10,7 +10,15 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- There is a new `Receiver.triggered` method that can be used instead of `selected_from`:
+
+  ```python
+  async for selected in select(recv1, recv2):
+      if recv1.triggered(selected):
+          print('Received from recv1:', selected.message)
+      if recv2.triggered(selected):
+          print('Received from recv2:', selected.message)
+  ```
 
 ## Bug Fixes
 
--- a/src/frequenz/channels/_receiver.py
+++ b/src/frequenz/channels/_receiver.py
@@ -373,9 +373,7 @@ class _Mapper(
             ReceiverStoppedError: If the receiver stopped producing messages.
             ReceiverError: If there is a problem with the receiver.
         """
-        return self._mapping_function(
-            self._receiver.consume()
-        )  # pylint: disable=protected-access
+        return self._mapping_function(self._receiver.consume())
 
     def __str__(self) -> str:
         """Return a string representation of the mapper."""
--- a/src/frequenz/channels/_receiver.py
+++ b/src/frequenz/channels/_receiver.py
@@ -155,10 +155,13 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import Generic, Self
+from typing import TYPE_CHECKING, Any, Generic, Self, TypeGuard
 
 from ._exceptions import Error
 from ._generic import MappedMessageT_co, ReceiverMessageT_co
+
+if TYPE_CHECKING:
+    from ._select import Selected
 
 
 class Receiver(ABC, Generic[ReceiverMessageT_co]):
@@ -283,6 +286,30 @@ class Receiver(ABC, Generic[ReceiverMessageT_co]):
             A new receiver that only receives messages that pass the filter.
         """
         return _Filter(receiver=self, filter_function=filter_function)
+
+    def triggered(
+        self, selected: Selected[Any]
+    ) -> TypeGuard[Selected[ReceiverMessageT_co]]:
+        """Check whether this receiver was selected by [`select()`][frequenz.channels.select].
+
+        This method is used in conjunction with the
+        [`Selected`][frequenz.channels.Selected] class to determine which receiver was
+        selected in `select()` iteration.
+
+        It also works as a [type guard][typing.TypeGuard] to narrow the type of the
+        `Selected` instance to the type of the receiver.
+
+        Please see [`select()`][frequenz.channels.select] for an example.
+
+        Args:
+            selected: The result of a `select()` iteration.
+
+        Returns:
+            Whether this receiver was selected.
+        """
+        if handled := selected._recv is self:  # pylint: disable=protected-access
+            selected._handled = True  # pylint: disable=protected-access
+        return handled
 
 
 class ReceiverError(Error, Generic[ReceiverMessageT_co]):
--- a/src/frequenz/channels/_select.py
+++ b/src/frequenz/channels/_select.py
@@ -376,14 +376,14 @@ async def select(  # noqa: DOC503
         import datetime
         from typing import assert_never
 
-        from frequenz.channels import ReceiverStoppedError, select, selected_from
+        from frequenz.channels import ReceiverStoppedError, select
         from frequenz.channels.timer import SkipMissedAndDrift, Timer, TriggerAllMissed
 
         timer1 = Timer(datetime.timedelta(seconds=1), TriggerAllMissed())
         timer2 = Timer(datetime.timedelta(seconds=0.5), SkipMissedAndDrift())
 
         async for selected in select(timer1, timer2):
-            if selected_from(selected, timer1):
+            if timer1.triggered(selected):
                 # Beware: `selected.message` might raise an exception, you can always
                 # check for exceptions with `selected.exception` first or use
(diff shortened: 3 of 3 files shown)


shsms:
Other naming options: `matches`, `matches_selected`

src/frequenz/channels/_receiver.py
```diff
@@ -155,11 +155,14 @@
...
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import Generic, Self
+from typing import TYPE_CHECKING, Any, Generic, Self, TypeGuard
 
 from ._exceptions import Error
 from ._generic import MappedMessageT_co, ReceiverMessageT_co
 
+if TYPE_CHECKING:
+    from ._select import Selected
```
llucax, line 164: :unamused: 

Oh, this is also an ugly side-effect of moving part of the implementation of `select` to `Receiver`. We could consider moving `Selected` to this file, if we are mixing stuff up anyway.

llucax (commented):
OK, so the issue we had with `TypeGuard` as a member method was not it being a member method but about [applying the type guard to `self`](https://github.com/microsoft/pyright/discussions/3125), this is why we can't do `selected.source(receiver)`, because we want to narrow `selected`, not `receiver`, but when used as a member the variable to narrow is the second argument, not `self`.

So `receiver.is_selected(selected)` works and will narrow `selected. OK. I don't like the fact that now part of `select()` is implemented in `Receiver` unless we have a considerable improvement in how the code is read, and `receiver.is_selected(selected)` doesn't read much better than `selected_from(selected, receiver)`.

But maybe something like `receiver.is_source_of(selected)` could work. What do you think?

src/frequenz/channels/_receiver.py
shsms: True, maybe also rename it so it sounds like `WrappedMessage` or `ReceivedMessage`, and it would be decoupled from `select` completely?

If not, it feels a bit weird, defining a `select` feature in the receiver module.

llucax: I don't know, for me it is unfortunate that we need to implement part of `select()` in receiver but it doesn't make sense to use weirder names only to try to hide that. `Selected` / `WrappedMessage` / `ReceivedMessage` will only be used by select anyways, I would keep calling it `Selected`.

I really have double feelings about this PR, for one side I think it improves the UX slightly, but I'm not sure if enough to worth a deprecation and pushing people to update their code, and for the other side I think it makes the implementation slightly less clean (more coupling between `Receiver` and `select`), but I'm not sure if enough to completely oppose to the change :smile: :disappointed:

shsms: > I think it makes the implementation slightly less clean (more coupling between `Receiver` and `select`)

I would look at it as an increased commitment to having `select` as a core component of the channels package, select is very deeply related to receivers.

And if it was a different language, we wouldn't even notice the coupling, because no language makes so much noise about these things as python.

shsms: > I think it improves the UX slightly,

I think it is more than a slight improvement for new users, because they don't have to import the weird-looking `selected_from` separately, so there's less confusion.

Marenz: I think we can support the the old way for long to stay backwards compatible to avoid annoying users though?

shsms: Oh right, so we can't rename `Selected`.

Also you mean add the new stuff, but don't deprecate the `selected_from` method?

llucax: Yeah, maybe we can deprecate it after a few more releases, when the new form is more widely used, to confirm it is really better than the old form.

llucax: > Oh right, so we can't rename `Selected`.

We can always rename it and create a (deprecated) alias using the old name. But personally I wouldn't do it, I think that name is fine.

Anyway, as I said, my main complain about this PR is the current `is_selected()` name, if we get a better name I'm OK with merging. I think my preferred name is `is_source_of()` but I can also live with `matches()`

llucax: About names, I must admit I also thought about `was`, which sucks as a name taken out of context, but reads very nicely when put in `receiver.was(selected)` :smile: :grimacing: 
ChatGPT also suggested `receiver.is_active(selected)` which doesn't reads as nicely as `receiver.is_actively(selected)` but makes much more sense as a function name.

EDIT: One issue with `is_active()` is it sounds much less obvious that you should always call it, even if you then ignore the receiver, but I guess that's true for most options, but I think it is better captured by `selected_from` or  `receiver.was(selected)`. I guess `matches` is also better than `is_active` or `is_source_of` in this sense. Since this is a very weird construct anyways, I'm really considering `receiver.was(selected)` as an option...

shsms: Similar to `was`, but maybe a bit more obvious, and independent of the variable naming conventions for `selected`, maybe `receiver.produced(selected)` is an option.  `was` doesn't work IMO, because it depends on the variable name `selected`, which we don't have any control over.

But I'm still in favour of `matches`, because it has the coolest sound, with a lispy/functional feel, in a select block:

```python
async for selected in select(power_rx, timer):
    if power_rx.matches(selected):
        print(f"Power: {power}")
    elif timer.matches(selected):
        print("Timer triggered")
```

llucax: In the line of `produced`, what about `triggered`?

```py
async for selected in select(power_rx, timer):
    if power_rx.triggered(selected):
        print(f"Power: {power}")
    elif timer.triggered(selected):
        print("Timer triggered")
```

What it bothers me about `matches` is it is really the `selected` that matches the receiver, not the other way around (if we could use `selected.matches(receiver)` that would be a :100: ). But as I said, I can live with `matches`.

shsms: You were right.  I'm also having doubts about `matches`.  I will update the PR to say `triggered`.

llucax: :scream:

RELEASE_NOTES.md
```diff
@@ -10,7 +10,15 @@
...
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- There is a new `Receiver.triggered` method that can be used instead of `selected_from`:
+
+  ```python
+  async for selected in select(recv1, recv2):
+      if recv1.triggered(selected):
+          print('Received from recv1:', selected.value)
+      if recv2.triggered(selected):
+          print('Received from recv2:', selected.value)
```
```suggestion
          print('Received from recv1:', selected.message)
      if recv2.triggered(selected):
          print('Received from recv2:', selected.message)
```

src/frequenz/channels/_receiver.py
```diff
@@ -373,9 +400,7 @@ def consume(self) -> MappedMessageT_co:  # noqa: DOC502
             ReceiverStoppedError: If the receiver stopped producing messages.
             ReceiverError: If there is a problem with the receiver.
         """
-        return self._mapping_function(
-            self._receiver.consume()
-        )  # pylint: disable=protected-access
+        return self._mapping_function(self._receiver.consume())
```
llucax, line 403: Optional: This is unrelated to the commit, could be moved to another commit.

src/frequenz/channels/_select.py
```diff
@@ -378,7 +376,7 @@ async def select(  # noqa: DOC503
         import datetime
         from typing import assert_never
 
-        from frequenz.channels import ReceiverStoppedError, select, selected_from
```
llucax, line 381: This shouldn't be removed, or the example needs to be updated to use `Receiver.triggered()` instead.

RELEASE_NOTES.md
Marenz: Shouldn't sybil catch this? Ah, probably because it's a RELEASE_NOTES.md file.

llucax: We are not "testing" the `RELEASE_NOTES.md`. I think we discussed it at some point, my feeling is no, the release notes sometimes have also old code as example of how to migrate some code, that it is not intended to work. I think it would be difficult to test examples in the release notes.

shsms: fixed.

src/frequenz/channels/_receiver.py
shsms: done.

src/frequenz/channels/_select.py
shsms: oops, replaced with `Receiver.triggered()`

llucax (approved):

PR #329, opened 2024-10-29
Support for disabling overflow-warning logs in broadcast receivers

shsms:


llucax (commented):
Not super excited about the extra complexity, and I was thinking it would be great to manage this via the config system, but even if we have a separate logger for these warnings so it can be disabled more easily, this doesn't allow us to show warnings for some receivers but not others, so I guess it doesn't cut it.

My main concern is I guess we are adding this only to cope with the fact that we can't unsubscribe/remove receivers, right? Or am I missing any other use case for this? It seems like having a channel overflowing is not good unless you don´t really care about receiving data from it.

shsms:
It is for cases where we want just the latest value, but also want a trigger, want to put it in a `select`, etc.

llucax:
OK, and in that case it will be combined with `size=1` I guess. Maybe it is worth creating a different receiver for this use case? Maybe `new_xxx_receiver()` where I have no idea what *xxx* could be :laughing:

shsms:
How about a private method `Receiver._allow_overflows`, while we decide the public interface?

llucax:
Unless having this is urgent, I'd rather spend some time thinking about a final solution than introducing a hack. And if we are going for the hack, I would be more inclined to have `_warn_on_overflow: bool` as an argument to `new_receiver()` and mention in the docs that this is experimental. But I would prefer not to start making every feature experimental from now on either :laughing:

llucax:
Moving the discussion to:

* https://github.com/frequenz-floss/frequenz-channels-python/discussions/339

To avoid polluting the PR too much.

llucax (approved):

PR #332, opened 2024-11-14
Handle type guards properly in `Receiver.filter()`

llucax:
Now the `Receiver` type returned by `Receiver.filter()` will have the narrowed type when a `TypeGuard` is used.

### DIFF
--- a/src/frequenz/channels/_receiver.py
+++ b/src/frequenz/channels/_receiver.py
@@ -155,13 +155,16 @@ from __future__ import annotations
 
 from abc import ABC, abstractmethod
 from collections.abc import Callable
-from typing import TYPE_CHECKING, Any, Generic, Self, TypeGuard
+from typing import TYPE_CHECKING, Any, Generic, Self, TypeGuard, TypeVar, overload
 
 from ._exceptions import Error
 from ._generic import MappedMessageT_co, ReceiverMessageT_co
 
 if TYPE_CHECKING:
     from ._select import Selected
+
+FilteredMessageT_co = TypeVar("FilteredMessageT_co", covariant=True)
+"""Type variable for the filtered message type."""
 
 
 class Receiver(ABC, Generic[ReceiverMessageT_co]):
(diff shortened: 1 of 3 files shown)


src/frequenz/channels/_receiver.py
```diff
@@ -267,11 +270,67 @@ def map(
...
+
+    # We need to use Any here because otherwise _Filter would have to deal with two
+    # different signatures. We can create two filter classes, one for regular functions
+    # and one for type guards, but then there is no way to tell at runtime which
+    # function is a type guard and which isn't to instantiate the correct class.
+    # Using Any here has no impact though, as thanks to the overloads, only the
+    # overloaded types will be accepted.
+    def filter(
+        self, filter_function: Callable[[ReceiverMessageT_co], Any], /
+    ) -> Receiver[Any]:
```
shsms: You don't need to change `_Filter` because `TypeGuard` is just another name for bool, with some additional semantic meaning.  The type system knows assigning a `TypeGuard` value to a `bool` is fine.

```suggestion
    def filter(
        self,
        filter_function: (
            Callable[[ReceiverMessageT_co], bool]
            | Callable[[ReceiverMessageT_co], TypeGuard[FilteredMessageT_co]]
        ),
        /,
    ) -> Receiver[ReceiverMessageT_co] | Receiver[FilteredMessageT_co]:
```

llucax: Interesting, I think what I tried was: `filter_function: Callable[[ReceiverMessageT_co], bool | TypeGuard[FilteredMessageT_co]]` and that didn't work.

llucax:
Updated.

tests/test_broadcast.py
```diff
@@ -248,6 +249,29 @@ async def test_broadcast_filter() -> None:
...
+    receiver = chan.new_receiver().filter(_is_int)
+
+    await sender.send("hello")
+    await sender.send(8)
+
+    message = await receiver.receive()
+    assert message == 8
+    match message:
+        case int():
+            assert message == 8
```
ela-kotulska-frequenz: Why is it necessary to have this check twice? (in line 267 and 270 ? ` assert message == 8` )

ela-kotulska-frequenz (commented):
LGTM, this is great!

tests/test_broadcast.py
llucax: I guess it could just be a `pass` inside the `case`, true. I just needed to put something :D

llucax: I just changed the test to make sure the `case int()` is executed. `mypy` should verify this too thannks to the `assert_never`, but just in case.

llucax:
Updated again.

tests/test_broadcast.py
ela-kotulska-frequenz: Thanks! Now it is clear for me why it was necessary :)

tests/test_broadcast.py
```diff
@@ -248,6 +249,31 @@ async def test_broadcast_filter() -> None:
...
+async def test_broadcast_filter_type_guard() -> None:
+    """Ensure filter type guard works."""
+    chan = Broadcast[int | str](name="input-chan")
+    sender = chan.new_sender()
+
+    def _is_int(num: int | str) -> TypeGuard[int]:
+        return isinstance(num, int)
+
+    # filter out all numbers less than 10.
+    receiver = chan.new_receiver().filter(_is_int)
```
shsms, line 261: the comment needs to be updated, but looks great otherwise!

llucax: Done!

shsms (approved):
🎉

PR #340, opened 2024-11-21
Change FileWatcher to stop printing watchfiles debug log

ela-kotulska-frequenz:
The underlying Rust code performs periodic checks to see
if it should stop watching. These checks raise TimeoutError,
which by default would log misleading debug messages.
The debug log was printed every 5 sec on linux.
Setting yield_on_timeout=True makes it yield an empty
set instead of printing debug logs.
We handle these empty sets in the `ready()` method by
continuing to wait.

### DIFF
--- a/src/frequenz/channels/file_watcher.py
+++ b/src/frequenz/channels/file_watcher.py
@@ -159,6 +159,11 @@ class FileWatcher(Receiver[Event]):
             watch_filter=self._filter_events,
             force_polling=force_polling,
             poll_delay_ms=int(polling_interval.total_seconds() * 1_000),
+            # The underlying Rust code performs periodic checks to see if it should stop watching.
+            # These checks raise TimeoutError, which by default would log misleading debug messages.
+            # Setting yield_on_timeout=True makes it yield an empty set instead of printing debug
+            # logs. We handle these empty sets in the `ready()` method by continuing to wait.
+            yield_on_timeout=True,
         )
         self._awatch_stopped_exc: Exception | None = None
         self._changes: set[FileChange] = set()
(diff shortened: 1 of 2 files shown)


ela-kotulska-frequenz:
If we use or plan to use any Windows system, then this PR won't work.

src/frequenz/channels/file_watcher.py
```diff
@@ -159,6 +160,10 @@ def __init__(
             watch_filter=self._filter_events,
             force_polling=force_polling,
             poll_delay_ms=int(polling_interval.total_seconds() * 1_000),
+            # Two arguments below are ok for non-Windows systems.
+            # For more details, refer awatch documentation.
+            rust_timeout=sys.maxsize - 1,
+            yield_on_timeout=True,
```
shsms, line 166: Do you know what the impact of this would be on windows systems?

ela-kotulska-frequenz: https://github.com/samuelcolvin/watchfiles/issues/110

shsms:
> If we use or plan to use any Windows system, then this PR won't work.

We will also have to support Windows machines because there are existing users.

src/frequenz/channels/file_watcher.py
ela-kotulska-frequenz: But I think that if we change `rust_timeout=sys.maxsize - 1,` back to None.
Then it should work on Windows and won't print debug log. However I didn't test it.

This is because the timeout mechanism in awatch looks like this:
```
            if raw_changes == 'timeout':
                if yield_on_timeout:
                    yield set()
                else:
                    logger.debug('rust notify timeout, continuing')
```
So instead of printing debug log, it can yield set() and we can ignore it. 
Only then debug logs won't print if it awatch was killed or not.

llucax:
Isn't there any way to control DEBUG logging in watchfiles? For me that would be the best choice, this is just a DEBUG message!

llucax (commented):
I really think we shouldn't go for this, we just need to disable the DEBUG logging if we have DEBUG logging enabled but find these particular messages too annoying and unhelpful.

ela-kotulska-frequenz:
We decided to abort this PR and handle it higher level.

PR #347, opened 2024-11-29
Use a `float` for the tolerance in the timer tests

llucax:
Hopefully this finally fixes the flaky hypothesis tests.

- **Remove deprecated uses of pytest-asyncio**
- **Use a `float` for the tolerance in the timer tests**
- **Revert "Use less extreme values for min and max timedelta in tests"**

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -148,6 +148,7 @@ disable = [
 [tool.pytest.ini_options]
 testpaths = ["tests", "src"]
 asyncio_mode = "auto"
+asyncio_default_fixture_loop_scope = "function"
 required_plugins = ["pytest-asyncio", "pytest-mock"]
 markers = [
   "integration: integration tests (deselect with '-m \"not integration\"')",
--- a/tests/test_timer.py
+++ b/tests/test_timer.py
@@ -6,7 +6,6 @@
 
 import asyncio
 import enum
-from collections.abc import Iterator
 from datetime import timedelta
 
 import async_solipsism
@@ -22,13 +21,10 @@ from frequenz.channels.timer import (
 )
 
 
-# Setting 'autouse' has no effect as this method replaces the event loop for all tests in the file.
-@pytest.fixture()
-def event_loop() -> Iterator[async_solipsism.EventLoop]:
-    """Replace the loop with one that doesn't interact with the outside world."""
-    loop = async_solipsism.EventLoop()
-    yield loop
-    loop.close()
+@pytest.fixture(autouse=True)
+def event_loop_policy() -> async_solipsism.EventLoopPolicy:
+    """Return an event loop policy that uses the async solipsism event loop."""
+    return async_solipsism.EventLoopPolicy()
 
 
 # We give some extra room (dividing by 10) to the max and min to avoid flaky errors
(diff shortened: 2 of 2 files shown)


llucax:
Tests failed again when queuing a PR, so here is another attempt.

daniel-zullo-frequenz:
> Tests failed again when queuing a PR, so here is another attempt.

Hypothesis usually tells you how to reproduce the error  by temporarily adding @reproduce_failure({PARAMETERS, FOR, THE, TEST, HERE}) as a decorator on the test case. Have you tried that to be sure the patch solves the issue?

llucax:
Good tip. I actually validated it manually, but for a previous attempt and forgot to do it with the new approach. I will check with the decorator :+1: :100:

daniel-zullo-frequenz (dismissed):

llucax:
FYI, this was the failure:

```
______________________ test_policy_skip_missed_and_drift _______________________

    @hypothesis.given(
>       tolerance=st.integers(min_value=0, max_value=_max_timedelta_microseconds),
        **_calculate_next_tick_time_args,
    )

tests/test_timer.py:148: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

tolerance = 171726190479152817, now = 171726190479152817
scheduled_tick_time = -1, interval = 1

    @hypothesis.given(
        tolerance=st.integers(min_value=0, max_value=_max_timedelta_microseconds),
        **_calculate_next_tick_time_args,
    )
    def test_policy_skip_missed_and_drift(
        tolerance: int, now: int, scheduled_tick_time: int, interval: int
    ) -> None:
        """Test the SkipMissedAndDrift policy."""
        hypothesis.assume(now >= scheduled_tick_time)
    
        next_tick_time = SkipMissedAndDrift(
            delay_tolerance=timedelta(microseconds=tolerance)
        ).calculate_next_tick_time(
            now=now, interval=interval, scheduled_tick_time=scheduled_tick_time
        )
        if tolerance < interval:
            assert next_tick_time > now
        drift = now - scheduled_tick_time
        if drift > tolerance:
>           assert next_tick_time == now + interval
E           assert 0 == (171726190479152817 + 1)
E           Falsifying example: test_policy_skip_missed_and_drift(
E               tolerance=171_726_190_479_152_817,
E               now=171_726_190_479_152_817,
E               scheduled_tick_time=-1,
E               interval=1,  # or any other generated value
E           )

tests/test_timer.py:166: AssertionError
```

daniel-zullo-frequenz:
> FYI, this was the failure

I see in this case hypothesis only mentioned how to reproduce the error/falsify the example.

llucax:
Yeah, I don't know how `@reproduce_failure` is used exactly, but [`example()`](https://hypothesis.readthedocs.io/en/latest/reproducing.html#providing-explicit-examples) works pretty well. Adding some examples, although I can't test the falsifying example precisely because we switched from `int` to `float` for the problematic value.

With this in mind, it seems like it fixes the issue. Pushed some updates, adding the examples so they are always tested just in case.

daniel-zullo-frequenz (approved):

PR #348, opened 2024-11-29
Add a `Receiver.close()` method

### LINKED ISSUE #297: A way to explicitly stop a `Receiver`
If you stop reading from a channel, you will get an overflow warning message after a short while. 

Sometimes, however we want to properly stop and not get spammed with log warnings, so a convenient and explicit way to do that would be nice.

shsms:
This method would allow individual receivers to be closed, without affecting the underlying channel, if there is one.

### DIFF
--- a/src/frequenz/channels/_anycast.py
+++ b/src/frequenz/channels/_anycast.py
@@ -391,6 +391,9 @@ class _Receiver(Receiver[_T]):
         self._channel: Anycast[_T] = channel
         """The channel that this receiver belongs to."""
 
+        self._closed: bool = False
+        """Whether the receiver is closed."""
+
         self._next: _T | type[_Empty] = _Empty
 
     @override
@@ -408,6 +411,9 @@ class _Receiver(Receiver[_T]):
         # if a message is already ready, then return immediately.
         if self._next is not _Empty:
             return True
+
+        if self._closed:
+            return False
 
         # pylint: disable=protected-access
         while len(self._channel._deque) == 0:
@@ -436,6 +442,9 @@ class _Receiver(Receiver[_T]):
         ):
             raise ReceiverStoppedError(self) from ChannelClosedError(self._channel)
 
+        if self._next is _Empty and self._closed:
+            raise ReceiverStoppedError(self)
+
         assert (
             self._next is not _Empty
         ), "`consume()` must be preceded by a call to `ready()`"
@@ -445,6 +454,14 @@ class _Receiver(Receiver[_T]):
         self._next = _Empty
 
         return next_val
+
+    @override
+    def close(self) -> None:
+        """Close this receiver.
+
+        After closing, the receiver will not be able to receive any more messages.
+        """
+        self._closed = True
 
     def __str__(self) -> str:
         """Return a string representation of this receiver."""
--- a/src/frequenz/channels/_anycast.py
+++ b/src/frequenz/channels/_anycast.py
@@ -10,6 +10,8 @@ from asyncio import Condition
 from collections import deque
 from typing import Generic, TypeVar
 
+from typing_extensions import override
+
 from ._exceptions import ChannelClosedError
 from ._generic import ChannelMessageT
 from ._receiver import Receiver, ReceiverStoppedError
@@ -320,6 +322,7 @@ class _Sender(Sender[_T]):
         self._channel: Anycast[_T] = channel
         """The channel that this sender belongs to."""
 
+    @override
     async def send(self, message: _T, /) -> None:
         """Send a message across the channel.
 
@@ -390,6 +393,7 @@ class _Receiver(Receiver[_T]):
 
         self._next: _T | type[_Empty] = _Empty
 
+    @override
     async def ready(self) -> bool:
         """Wait until the receiver is ready with a message or an error.
 
@@ -417,6 +421,7 @@ class _Receiver(Receiver[_T]):
         # pylint: enable=protected-access
         return True
 
+    @override
     def consume(self) -> _T:
         """Return the latest message once `ready()` is complete.
 
--- a/src/frequenz/channels/_merge.py
+++ b/src/frequenz/channels/_merge.py
@@ -191,6 +191,21 @@ class Merger(Receiver[ReceiverMessageT_co]):
 
         return self._results.popleft()
 
+    @override
+    def close(self) -> None:
+        """Close the receiver.
+
+        After calling this method, new messages will not be received.  Once the
+        receiver's buffer is drained, trying to receive a message will raise a
+        [`ReceiverStoppedError`][frequenz.channels.ReceiverStoppedError].
+        """
+        for task in self._pending:
+            if not task.done() and task.get_loop().is_running():
+                task.cancel()
+        self._pending = set()
+        for recv in self._receivers.values():
+            recv.close()
+
     def __str__(self) -> str:
         """Return a string representation of this receiver."""
         if len(self._receivers) > 3:
--- a/src/frequenz/channels/_merge.py
+++ b/src/frequenz/channels/_merge.py
@@ -54,6 +54,8 @@ import itertools
 from collections import deque
 from typing import Any
 
+from typing_extensions import override
+
 from ._generic import ReceiverMessageT_co
 from ._receiver import Receiver, ReceiverStoppedError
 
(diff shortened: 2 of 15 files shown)


shsms:
Still need to add tests.

src/frequenz/channels/_merge.py
```diff
@@ -187,6 +191,21 @@ def consume(self) -> ReceiverMessageT_co:
...
+    @override
+    def close(self) -> None:
+        """Close the receiver.
+
+        After calling this method, new messages will not be received.  Once the
+        receiver's buffer is drained, trying to receive a message will raise a
+        [`ReceiverStoppedError`][frequenz.channels.ReceiverStoppedError].
+        """
+        for task in self._pending:
+            if not task.done() and task.get_loop().is_running():
```
llucax, line 203: Mmm, interesting the `task.get_loop().is_running()`. Was this to address warnings in tests or any other reason? I wonder if it is a good practice to do in general or not (and for a normal cleanup all tasks should be cancelled before stopping the loop).

llucax (commented):
LGTM but the CI is failing, and this reminds me that it would be good to add tests at least for the receivers that actually do something when `close()` is called.

And oh, yes! As usual names, names, names :laughing: 

TL;DR: All good the name actually, just some other comments about future directions.

I was going to suggest to rename `close()` to `stop()`, for once to match the `ReceiverStoppedError` but also to match the upcoming [`Service` interface](https://github.com/frequenz-floss/frequenz-core-python/blob/v1.x.x/src/frequenz/core/asyncio/_service.py).

But I see close actually wait for the current buffer to be consumed, my idea for `stop()` would be that stuff gets just dropped.

Maybe to make it play better with `Service` we can keep `close()` as you implemented it, as a "graceful" shutdown, and when we introduce `Service` we can use `cancel()` to drop the current buffers. Then maybe `stop()` can take an option to decide if `close()` or `cancel()` is called (stop is usually implemented as `self.cancel(); await self`).

shsms:
> LGTM but the CI is failing, and this reminds me that it would be good to add tests at least for the receivers that actually do something when `close()` is called.

Yes, I'm already working on tests in this same branch, just made an early PR because the code is ready for review.

> Maybe to make it play better with `Service` we can keep `close()` as you implemented it, as a "graceful" shutdown, and when we introduce `Service` we can use `cancel()` to drop the current buffers. Then maybe `stop()` can take an option to decide if `close()` or `cancel()` is called (stop is usually implemented as `self.cancel(); await self`).

I think `stop()` doesn't make sense for receivers, because it is not a task.  Idiomatically, we should have `stop()` only for things that look like there must be an activity there.  So it is great for `Actor`, etc.  But for a stream, I like `close()` better.

src/frequenz/channels/_merge.py
shsms: > Was this to address warnings in tests or any other reason?

Yes, I think I started adding it just for tests and I'm sure we'll find it in other places in this repo as well (unless you removed it).

> I wonder if it is a good practice to do in general or not (and for a normal cleanup all tasks should be cancelled before stopping the loop).

I agree, these could go away once we reach a point where removing them won't cause warning spam.

llucax:
> I think stop() doesn't make sense for receivers, because it is not a task. Idiomatically, we should have stop() only for things that look like there must be an activity there. So it is great for Actor, etc. But for a stream, I like close() better.

Mmm, receiver don't **need** to have task, but many **do** (like `Merger`, `Timer` and `FileWatcher`). If we don't add a task-like interface, it means we can't stop these receivers using the `Receiver` interface, which is the only way to access them if we apply a `filter` or `map` to them for example.

But maybe there are other solutions to this. We can discuss this when the time comes.

llucax:
Please let me know when this is ready for a final review :)

shsms:
> Mmm, receiver don't **need** to have task, but many **do** (like `Merger`, `Timer` and `FileWatcher`). If we don't add a task-like interface, it means we can't stop these receivers using the `Receiver` interface, which is the only way to access them if we apply a `filter` or `map` to them for example.

That doesn't sound like an issue, because people still call `close()` on the receiver, and if the underlying thing is a `FileWatcher` for example, the `close()` would just call the `stop()`, which is what this PR is doing already.

And calling `close()` on `Map` and `Filter` would call `close()` of the underlying receiver.  This PR has that too.

llucax:
> That doesn't sound like an issue, because people still call close() on the receiver, and if the underlying thing is a FileWatcher for example, the close() would just call the stop(), which is what this PR is doing already.

But `stop()` is async, and it is important to be able to wait until the thing actually stopped to avoid those annoying warnings in test we were talking about :laughing: 

If we make `close()` async, then fine, but then we are ACKing that receivers can have background running tasks, and then why shouldn't they follow the `Service` interface?

shsms:
This is ready, the only big change is that `Receiver.close()` is now async.

llucax:
Last push to have a (`Background`)`Service`-compatible interface.

> If we make `close()` async, then fine, but then we are ACKing that receivers can have background running tasks, and then why shouldn't they follow the `Service` interface?

llucax:
BTW, the method can't be abstract, otherwise this is a breaking change. We need to provide a default empty implementation I guess. We can create an issue to make the method abstract for 2.0.

src/frequenz/channels/file_watcher.py
```diff
@@ -229,6 +232,11 @@ def consume(self) -> Event:
         change, path_str = self._changes.pop()
         return Event(type=EventType(change), path=pathlib.Path(path_str))
 
+    @override
+    async def close(self) -> None:
+        """Close this receiver."""
+        self._stop_event.set()
```
llucax, line 238: We should probably do a `await self._awatch.aclose()` here to make sure all the threads from the awatch are stopped too. Maybe even mark the receiver somehow as stopped so `ready()` doesn ´t try to use `self._awatch` again? Not sure what would happen otherwise, what kind of exception would be raised.

src/frequenz/channels/_anycast.py
```diff
@@ -441,6 +455,14 @@ def consume(self) -> _T:
 
         return next_val
 
+    @override
+    async def close(self) -> None:
+        """Close this receiver.
+
+        After closing, the receiver will not be able to receive any more messages.
+        """
+        self._closed = True
```
llucax, line 464: I think here we should also break the `ready()` if it is waiting for the condition variable. I guess this would mean adding a an event and then `async.wait()` on the first one to complete in `ready()`?

I guess this is actually the difference between *closing* and *stopping*. Close is basically analogous to `Task.cancel()`, it is like it schedules the stopping. And it is actually probably a good idea to have it separated, but we are still missing the `await` then, making `stop()` async but then not actually awaiting for anything seems weird to me.

llucax (commented):
I didn't review everything, because I think the review will change a lot depending on what we do with `close` vs `stop()` (`stop()` being `close()`/`cancel()` + `__await__`).

`cancel()` and `__await__` are actually the 2 primitives used by (`Background`)`Service` too, based on `Task`, and the more I look at it, the more I think receivers should follow this interface too.

shsms:
I've made the `close()` method not abstract and not async.  With these,
- it is no longer  a breaking change
- it becomes a signal to close(), in keeping with the idioms of the asyncio library, which was the behaviour already even though it was async, and not a "ensure it is closed before returning".

With this the two above concerns should also go away.

src/frequenz/channels/file_watcher.py
shsms: switched back to non-async `close()`.

src/frequenz/channels/_anycast.py
shsms: switched back to non-async `close()`.

shsms:
Rebased on latest

shsms:
@llucax could you take a look at this again?

src/frequenz/channels/_receiver.py
```diff
@@ -217,6 +217,15 @@ def consume(self) -> ReceiverMessageT_co:
...
         """
 
+    def close(self) -> None:
+        """Close the receiver.
+
+        After calling this method, new messages will not be available from the receiver.
+        Once the receiver's buffer is drained, trying to receive a message will raise a
+        [`ReceiverStoppedError`][frequenz.channels.ReceiverStoppedError].
+        """
+        raise NotImplementedError("close() must be implemented by subclasses")
```
llucax, line 227: So we don't forget:
* https://github.com/frequenz-floss/frequenz-channels-python/issues/365

llucax (approved):

PR #357, opened 2024-12-04
Rename `OnlyIfPrevious` to `WithPrevious`

### LINKED ISSUE #355: Rename `OnlyIfPrevious` as `WithPrevious`
### What's needed?

Given #354, there is no more need to make it more clear if we are accepting messages for which the predicate returns `True` or `False`.

### Proposed solution

Rename to `WithPrevious`.

### Use cases

_No response_

_No response_

### Additional context

_No response_

llucax:
The new name makes is more clear that this predicate is adding an extra condition to the message filtering based on the previous message.

Fixes #355.


src/frequenz/channels/experimental/_with_previous.py
```diff
@@ -1,7 +1,7 @@
 # License: MIT
 # Copyright © 2024 Frequenz Energy-as-a-Service GmbH
 
-"""Predicates to be used in conjuntion with `Receiver.filter()`."""
+"""Composable predicate to be cache and compare with the previous message."""
```
daniel-zullo-frequenz: ?
```suggestion
"""Composable predicate to cache and compare with the previous message."""
```

daniel-zullo-frequenz (commented):
Just a cosmetic about documentation, LGTM

src/frequenz/channels/experimental/_with_previous.py
llucax: Fixed, thanks!

daniel-zullo-frequenz (approved):

PR #368, opened 2025-01-21
Explain the `Receiver.close()` method in the user guide

### LINKED ISSUE #297: A way to explicitly stop a `Receiver`

shsms:


llucax (dismissed):

llucax:
Maybe add a more explicit mention about cleanup too (it sounds a bit too optional IMHO). Maybe making a parallel with files or sockets?

shsms:
I've added a fixup commit.  Does that look better?

llucax (approved):
Yup, LGTM. We can also use [`with closing(recv)`](https://docs.python.org/3/library/contextlib.html#contextlib.closing) :star_struck:

shsms:
Oops, need to squash

shsms:
Squashed.

Issue #365, opened 2025-01-16
Breaking changes for v2.0


llucax:
This is a tracking issue to document all breaking changes we want to introduce for v2.0.

Once we have a more clear plan to go for 2.0, we can convert some of the TODOs in their own issues to make tracking of different changes easier.

- [ ] Make `Receiver.close()` abstract.
- [ ] In `new_receiver()`, rename the argument `limit` to `buffer_size`.
- [ ] Rename 1.x `Timer` to `MonotonicTimer`.  Add a `Timer` that works with the wall clock.
- [ ] Add a `tick_on_init` parameter to `Timer` and change tick behaviour to the following:

    | tick_on_init | start_delay | tick at 0 | tick after delay |
    |--------------|-------------|-----------|------------------|
    | true         | > 0         | true      | true             |
    | true         | 0           | true      | -                |
    | false        | > 0         | false     | true             |
    | false        | 0           | false     | -                |

- [ ] Remove `channel.close()` method, which has been replaced by `channel.aclose()`

llucax:
When we have a wall clock timer, I feel the complexity of the monotonic timer can be greatly reduced, as many of the features needed there are just to support using it as a base to construct wall clock timers, so I hope in v2 we can actually remove complexity from the monotonic timer (like `start_delay`) instead of adding more.

PR #372, opened 2025-02-14
An optional `tick_at_start` parameter has been added to `Timer`

shsms:
When `True`, the timer will trigger immediately after starting, and then wait for the interval before triggering again.

### DIFF
--- a/src/frequenz/channels/timer.py
+++ b/src/frequenz/channels/timer.py
@@ -477,6 +477,7 @@ class Timer(Receiver[timedelta]):
         *,
         auto_start: bool = True,
         start_delay: timedelta = timedelta(0),
+        tick_at_start: bool = False,
         loop: asyncio.AbstractEventLoop | None = None,
     ) -> None:
         """Initialize this timer.
@@ -497,6 +498,13 @@ class Timer(Receiver[timedelta]):
             start_delay: The delay before the timer should start. If `auto_start` is
                 `False`, an exception is raised. This has microseconds resolution,
                 anything smaller than a microsecond means no delay.
+            tick_at_start: When `True`, the timer will trigger immediately after
+                starting, and then wait for the interval before triggering
+                again.  When `False`, the timer will wait the interval before
+                triggering for the first time.  If `auto_start` is `False` and
+                this is `True`, an exception is raised.  If a `start_delay` is
+                specified and this is `True`, the first trigger will be immediately
+                after the `start_delay`.
             loop: The event loop to use to track time. If `None`,
                 `asyncio.get_running_loop()` will be used.
 
@@ -516,6 +524,9 @@ class Timer(Receiver[timedelta]):
             raise ValueError(
                 "`auto_start` must be `True` if a `start_delay` is specified"
             )
+
+        if tick_at_start is True and auto_start is False:
+            raise ValueError("`auto_start` must be `True` if `tick_at_start` is `True`")
 
         self._interval: int = _to_microseconds(interval)
         """The time to between timer ticks."""
@@ -567,7 +578,7 @@ class Timer(Receiver[timedelta]):
         """
 
         if auto_start:
-            self.reset(start_delay=start_delay)
+            self.reset(start_delay=start_delay, tick_at_start=tick_at_start)
 
     @property
     def interval(self) -> timedelta:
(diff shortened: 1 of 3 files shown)


shsms:
FYI @cwasicki

src/frequenz/channels/timer.py
```diff
@@ -497,6 +498,13 @@ def __init__(  # noqa: DOC503 pylint: disable=too-many-arguments
             start_delay: The delay before the timer should start. If `auto_start` is
                 `False`, an exception is raised. This has microseconds resolution,
                 anything smaller than a microsecond means no delay.
+            tick_at_start: When `True`, the timer will trigger immediately after
+                starting, and then wait for the interval before triggering
+                again.  When `False`, the timer will wait the interval before
+                triggering for the first time.  If `auto_start` is `False` and
+                this is `True`, an exception is raised.  If a `start_delay` is
+                specified and this is `True`, the first trigger will be immediately
+                after the `start_delay`.
```
cwasicki, line 507: One use-case we have is that you want to trigger immediately, then wait for `start_delay` and from that point onward trigger each interval. This is not possible still with this addition, correct?

cwasicki: I wonder whether it should be `tick_on_init` and the combination of `tick_at_init=True` and `start_delay=X` should function like that?

shsms: > This is not possible still with this addition, correct?

No, I remember that use case and this should reduce its complexity and make it readable.

But I feel that this is the right balance, and `tick_on_init` + `tick_on_start` would make the API for the `Timer` too complex.

But we can come back to it.  And if we are going to add more `tick_on_*` policies, maybe we should consider an enum instead, describing when to tick.

cwasicki: What I meant is replace `tick_on_start` with `tick_on_init` because AFAIU the behavior here can be achieved with a `tick_on_init` but what I outlined above is not possible with a `tick_on_start`.

shsms: If we have only `tick_on_init`, then someone who wants just `tick_on_start` will get too many ticks and would need more complicated workarounds than someone who needs `tick_on_init` but only has `tick_on_start`.

cwasicki: Which case do you mean here?

cwasicki: Discussed in person, we take it in for now and keep potential changes for v2.0

cwasicki (approved):

src/frequenz/channels/timer.py
shsms: I've added an item describing the 2.0 plan here: https://github.com/frequenz-floss/frequenz-channels-python/issues/365

src/frequenz/channels/timer.py
```diff
@@ -497,6 +498,13 @@ def __init__(  # noqa: DOC503 pylint: disable=too-many-arguments
             start_delay: The delay before the timer should start. If `auto_start` is
                 `False`, an exception is raised. This has microseconds resolution,
                 anything smaller than a microsecond means no delay.
+            tick_at_start: When `True`, the timer will trigger immediately after
+                starting, and then wait for the interval before triggering
+                again.  When `False`, the timer will wait the interval before
+                triggering for the first time.  If `auto_start` is `False` and
+                this is `True`, an exception is raised.  If a `start_delay` is
```
llucax, line 505: I don't get why raising if `auto_start` is `False`? What is the issue with making this first tick delayed to when the timer actually starts, like when using `start_delay`?

llucax (commented):
LGTM otherwise.

src/frequenz/channels/timer.py
shsms: Restricted by how it is implemented.  But if they need it, users can specify `tick_on_start=True` in the call to `reset`, just like with `start_delay`.

shsms: > What is the issue with making this first tick delayed to when the timer actually starts, like when using `start_delay`?

No, that's not how `start_delay` is implemented.  It also doesn't work with `auto_start=False`.

llucax: Mmm, OK.

llucax (approved):

PR #383, opened 2025-03-26
Exclude repo-config from dependabot grouping

llucax:
This dependency usually need manual intervention to update, so it's better to exclude it from the grouping, which usually blocks merging until the manual intervention happens.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -19,17 +19,25 @@ updates:
     # Major updates are still managed, but they'll create one PR per
     # dependency, as major updates are expected to be breaking, it is better to
     # manage them individually.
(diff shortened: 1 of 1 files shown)


llucax:
I will force-merge this, as the diff is pretty trivial and I want to use it as a test for https://github.com/frequenz-floss/frequenz-repo-config-python/pull/368.

llucax:
I'm getting a :unicorn: when trying to force-merge, saying that there is already a "merge in progress" :shrug: 
So if anyone can approve to see if it works via the queue, it would be helpful.

llucax:
OK, it got de- :unicorn: ed by itself.

PR #385, opened 2025-03-26
Use a wildcard to exclude repo-config from dependabot groups

llucax:
It looks like dependabot considers optional dependencies as part of the dependency name, so `frequenz-repo-config[lib]` doesn't match the exclude rule for `frequenz-repo-config`.


llucax:
Also force-merging, hopefully this does the trick.

PR #389, opened 2025-03-27
Use a new grouping strategy for dependabot updates

llucax:
We group patch updates as they should always work.

We also group minor updates, as it works too for most libraries, typically except libraries that don't have a stable release yet (v0.x.x branch), so we make some exceptions for them.

Major updates and dependencies excluded by the above groups are still managed, but they'll create one PR per dependency, as breaking is expected so it might need manual intervention.

Finally we group some dependencies that are related to each other, and usually needs to be updated together.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -13,31 +13,42 @@ updates:
     versioning-strategy: auto
     # Allow up to 10 open pull requests for updates to dependency versions
     open-pull-requests-limit: 10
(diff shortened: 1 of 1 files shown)


llucax:
Will force-merge too, apparently it worked but it took it some time to get the new PRs out, but we need to group repo-config updates. I was also thinking that probably another strategy makes more sense, based on likelihood of breaking changes.

PR #397, opened 2025-03-27
Implement `fork` method in the `Receiver`

### LINKED ISSUE #382: Add a `Receiver.new_receiver()` method
### What's needed?

A way to, given a receiver, create a new receiver to receive from the same *channel* or *source*.

### Proposed solution

Add a `Receiver.new_receiver()` method that will basically create a new receiver in the same channel from the original receiver, with the same configuration of the receiver. It is basically a forward to `channel.new_receiver()`, when a channel even exists. The default implementation will be to forward all messages received by the original receiver to the new receiver too (implemented using `Pipe` probably). `Map` and `Filter` would not use a pipe, but just call `new_receiver` on their source receiver.

### Use cases

When working with actors, it is sometimes useful to split big tasks into smaller sub-actors. In this case, the parent actor needs to feed these sub-actors with receivers that comes from the receivers passed to the parent actor (for example when listening to dispatches, the subactors might need to listen to dispatches too).

In this case, the parent actor (which normally don't have access to the original channel) needs to create new channels to forward the messages to the sub-actors.

ela-kotulska-frequenz:
This is useful for scenarios where multiple consumers need to process the same stream of messages.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -11,6 +11,9 @@
 ## New Features
 
 - An optional `tick_at_start` parameter has been added to `Timer`.  When `True`, the timer will trigger immediately after starting, and then wait for the interval before triggering again.
+- Add `Receiver.fork` method to create independent clones of the receiver.
+    - Useful for scenarios where multiple consumers need to process the same stream of messages. Each forked receiver.
+    - Each forked receiver maintains its own independent message queue
 
 ## Bug Fixes
 
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -26,7 +26,7 @@ classifiers = [
 ]
 requires-python = ">= 3.11, < 4"
 dependencies = [
-  "typing-extensions >= 4.5.0, < 5",
+  "typing-extensions >= 4.11.0, < 5",
   "watchfiles >= 0.15.0, < 1.1.0",
 ]
 dynamic = ["version"]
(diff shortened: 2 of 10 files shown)


ela-kotulska-frequenz:
But I am not sure this is correct way, because `fork` for FileWatcher or Timer looks strange. And I don't see any use case for them.
If we agree to go with it, then I will add tests and improve the code.

shsms:
I didn't realize you were picking this already, I think the proposal needs some changes (sorry, I should have written something earlier).

`fork` doesn't tell if it should be a broadcast or anycast style behaviour, and users might need different things.  I would prefer using `experimental.Pipe` for this purpose instead because then,  we don't just inherit the behaviour of anycast/broadcast receivers, and don't create additional timers, filewatchers etc, but instead just broadcast their results.

llucax:
Same here, I just wrote it as a proposal, without thinking too much about it, just so we don't forget to discuss it but I didn't think you were going to take it so fast, I guess it probably needs a bit more discussion and thought. Maybe we can discuss in the next SDK meeting.

llucax:
BTW, from my side we didn't have to close this, only that it needed a bit more discussion. At least **I** think it could still be a valid approach.

llucax:
> `fork` doesn't tell if it should be a broadcast or anycast style behaviour, and users might need different things. I would prefer using `experimental.Pipe` for this purpose instead because then, we don't just inherit the behaviour of anycast/broadcast receivers, and don't create additional timers, filewatchers etc, but instead just broadcast their results.

BTW, the original proposal (I didn't check the code in the PR yet) suggested it is basically a proxy for `underlaying_source.new_receiver()` so if you fork a broadcast receiver, you get another broadcast receiver, if you fork a anycast receiver, you get a new anycast receiver. So I don't think this is ambiguous, at least as long as you know what type of receiver you are forking.

PR #408, opened 2025-04-09
Rename `close()` to `aclose()` in `Broadcast` and `Anycast`

### LINKED ISSUE #379: Add a new alias for channels `async close()` called `aclose()`
In Python there are already a few `async` close methods, like [for async generators](https://docs.python.org/3/reference/expressions.html#agen.aclose). It also includes a [`aclosing`](https://docs.python.org/3/library/contextlib.html#contextlib.aclosing) context manager to automatically close these objects.

Also mark `close()` as deprecated.

llucax:
This is done to follow Python's convention. With this change now channels can be used with the `aclosing()` context manager for example.

The name `close()` is still available for backwards compatibility, but deprecated.

Fixes #379.

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -7,6 +7,9 @@
 ## Upgrading
 
 - Some minimal dependencies have been bumped, so you might need to adjust your dependencies accordingly.
+- `Broadcast` and `Anycast` channels method `close()` was renamed to `aclose()` to follow Python's convention. With this change now channels can be used with the [`aclosing()`](https://docs.python.org/3/library/contextlib.html#contextlib.aclosing) context manager for example.
+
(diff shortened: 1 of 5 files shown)


shsms:
Wait, haven't we spoken about this _many_ times already?  The `close` method doesn't wait for the channel to drain completely and fully close.  It just stops the channel from accepting new messages, and initiates the close process, and so `aclose` doesn't make sense?

shsms:
Maybe we change the implementation to wait until the receivers are drained?  Alternatively, create a task that would trigger the condition variable, and make `close` sync.

llucax:
Moving the discussion to the issue (in case we close this PR in favor of another approach, so the discussion is preserved until the issue is fixed):

* https://github.com/frequenz-floss/frequenz-channels-python/issues/379#issuecomment-2789403674

shsms (approved):
I've changed my opinion on this and think the action we take (of stopping senders) is already sufficient to call this `aclose`.

Issue #116, opened 2023-06-09
Improve composability of receivers


llucax:
### What's needed?

It would be nice for `select()` to be more flexible (add and remove receivers, have optional receivers) without losing its static-ness.

### Proposed solution

Improve the composability of existing receivers and create a new one:

- [ ] Add a channel or receiver that merges messages from multiple receivers (or recycle `Merger` to do that) and can dynamically add and remove receivers, and just waits forever (it's never ready) when there are no receivers
- [x] Maybe add a wrapper on the above that only accepts 0 or 1 receivers, to implement an *optional receiver* (#409)
- [ ] Maybe add a `NopReceiver` that does the same as `OptionalReceiver(None)`

### Use cases

I think many cases where users need to handle lots of tasks manually to cope with more dynamic situations, like the resampler, could be greatly simplified if they could be handled automatically by using a more flexible select.

### Alternatives and workarounds

Break the select loop, use tasks manually.

### Additional context

This would fix a few open issues / discussions:

* https://github.com/frequenz-floss/frequenz-channels-python/discussions/24
* https://github.com/frequenz-floss/frequenz-channels-python/issues/12

llucax:
Maybe I should have made this a discussion first. If anyone sees any controversy in this, we can convert it to a discussion.  If there are no big objections, we can turn the TODO list into issues as we plan to address them.

shsms:
> I think many cases where users need to handle lots of tasks manually to cope with more dynamic situations, like the resampler, could be greatly simplified if they could be handled automatically by using a more flexible select.

I guess I'm just having trouble imagining what kind of possibilities would come out of these features.  It sounds like they're to deal with scenarios where channels that we read from close for some reason.  Is that accurate or are you thinking of other scenarios as well?

Is the resampler the only known use-case for these features?  I'm not very familiar with it, so maybe that would be just a good first test: it would be nice to see how they would simplify the resampler.  Would it help eliminate the two *helper layers?

So we just make a basic interface without implementation for the features proposed in this issue,  then see how it affects the resampler,  just to make sure we limit feature creep.  

Or maybe even a simpler version if that sounds like too much work.

llucax:
>  It sounds like they're to deal with scenarios where channels that we read from close for some reason.

[...]

PR #409, opened 2025-04-14
Add OptionalReceiver for handling unset receivers

ela-kotulska-frequenz:
Add a new OptionalReceiver class that wraps an optional underlying receiver, allowing for indefinite waiting when no receiver is set. This eliminates the need for explicit if-else branches to check if receiver is None.

### DIFF
--- a/src/frequenz/channels/experimental/_optional_receiver.py
+++ b/src/frequenz/channels/experimental/_optional_receiver.py
@@ -1,0 +1,79 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""A receiver that will wait indefinitely if there is no underlying receiver.
+
+The `OptionalReceiver` is useful when the underlying receiver is not set initially.
+Instead of making `if-else` branches to check if the receiver is set, you can use
+this receiver to wait indefinitely if it is not set.
+"""
+
+import asyncio
+
+from typing_extensions import override
+
+from frequenz.channels import Receiver, ReceiverError, ReceiverMessageT_co
+
+
+class OptionalReceiver(Receiver[ReceiverMessageT_co]):
+    """A receiver that will wait indefinitely if there is no underlying receiver.
+
+    This receiver is useful when the underlying receiver is not set initially.
+    Instead of making `if-else` branches to check if the receiver is set, you can use
+    this receiver to wait indefinitely if it is not set.
+    """
+
+    def __init__(self, receiver: Receiver[ReceiverMessageT_co] | None):
+        """Initialize this instance.
+
+        Args:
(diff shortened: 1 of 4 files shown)


ela-kotulska-frequenz:
I called it `NullableReceiver` because `OptionalReceiver` suggests that receiver can be set during runtime.
It also allows us to create class with name `OptionalReceiver` later if needed.

llucax (dismissed):
LGTM, but for me the name `NullableReceiver` suggests more that it can be updated after created, as a nullable type is exactly that, a type that can be null or something else at any moment. In Python `Optional` is an alias for `| None` so it actually has the same meaning. `Optional` is also familiar in Python and has a well known meaning (and there is no "null", so "nullable" sounds weird to me in Python).

I don't know, **I would keep it as `OptionalReceiver`**, but will not oppose to `Nullable` too strongly as it is in experimental and there are other opportunities to find a better name.

llucax:
This actually makes me think that maybe in the future it would be best to implement optional receivers directly in `select()`, because it is actually an artifact we need only to use with `select()`, I don't see any other use case for it that is not `select()`, at least at the moment, at least until we allow for setting to `None` (or not) at runtime, in that case I can see that it could be used as some sort of generic pausing mechanism for receivers. Anyway, this is a discussion for #116.

llucax (dismissed):

ela-kotulska-frequenz:
>  because it is actually an artifact we need only to use with select()

When main actor creates sub-actors, it needs to know if dispatch receiver was set or not :)
And this was missing in the code. I just added `is_set` method to the `OptionalReceiver`.

But this is true, when we will have it in the `select` then it won't be needed! Thanks :)

llucax:
Yeah, I mean if `select()` can accept `None` and just ignore `None` arguments, that would already make the trick, at least **for this particular use case**.

llucax:
I think we should move forward with this for now, changing `select()` would take some time I guess, as it is probably more controversial.

llucax (approved):

src/frequenz/channels/experimental/_optional_receiver.py
```diff
@@ -0,0 +1,79 @@
...
+        return self._receiver.consume()
+
+    def is_set(self) -> bool:
+        """Check if the receiver is set."""
+        return self._receiver is not None
+
+    def close(self) -> None:
+        """Stop the receiver."""
+        if self._receiver is not None:
+            self._receiver.close()
```
shsms, line 79: If there is no receiver, you might want to keep a `closed` flag.  You can then return immediately when `ready` is called, and on `consume,` you can raise a `ReceiverClosedError`, like the other receivers.

PR #419, opened 2025-05-14
Implement a `NopReceiver`

shsms:
It is useful as a place-holder receiver for use in contexts where a receiver is necessary, but one is not available.

### DIFF
--- a/src/frequenz/channels/experimental/_nop_receiver.py
+++ b/src/frequenz/channels/experimental/_nop_receiver.py
@@ -1,0 +1,57 @@
+# License: MIT
+# Copyright © 2025 Frequenz Energy-as-a-Service GmbH
+
+"""A receiver that will never receive a message.
+
+It is useful as a place-holder receiver for use in contexts where a receiver is
+necessary, but one is not available.
+"""
+
+import asyncio
+
+from typing_extensions import override
+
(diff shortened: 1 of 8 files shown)


shsms:
Can drop the deprecation commit, if you want to postpone it.

src/frequenz/channels/experimental/_nop_receiver.py
```diff
@@ -0,0 +1,57 @@
...
+class NopReceiver(Receiver[ReceiverMessageT_co]):
+    """A place-holder receiver that will never receive a message."""
+
+    def __init__(self) -> None:
+        """Initialize this instance."""
+        self._closed: bool = False
+
+    @override
+    async def ready(self) -> bool:
+        """Wait for ever unless the receiver is closed.
```
Marenz, line 27: forever

Marenz (approved):

llucax (commented):
I know I suggested this name, but recently I needed something similar for a context manager and bumped into [`nullcontext`](https://docs.python.org/3/library/contextlib.html#contextlib.nullcontext), so maybe `NullReceiver` could be more familiar to python users.

But just an idea, I think the gain, if any, is marginal anyways...

PR #424, opened 2025-06-03
Support grouping by keys in `LatestValueCache`

shsms:
`LatestValueCache` now takes an optional `key` function.  When
specified, it is used to get the key for each incoming message, and
the latest value for each key is cached and can be retrieved
separately.

### DIFF
--- a/src/frequenz/channels/_latest_value_cache.py
+++ b/src/frequenz/channels/_latest_value_cache.py
@@ -8,11 +8,19 @@ long as there has been one value received.
 
 [LatestValueCache][frequenz.channels.LatestValueCache] takes a
 [Receiver][frequenz.channels.Receiver] as an argument and stores the latest
-value received by that receiver.  As soon as a value is received, its
+value received by that receiver.  It also takes an optional `key` function
+that allows you to group the values by a specific key.  If the `key` is
+provided, the cache will store the latest value for each key separately,
+otherwise it will store only the latest value received overall.
+
+As soon as a value is received, its
 [`has_value`][frequenz.channels.LatestValueCache.has_value] method returns
 `True`, and its [`get`][frequenz.channels.LatestValueCache.get] method returns
 the latest value received.  The `get` method will raise an exception if called
 before any messages have been received from the receiver.
+
+Both `has_value` and `get` methods can take an optional `key` argument to
+check or retrieve the latest value for that specific key.
 
 Example:
 ```python
@@ -32,12 +40,15 @@ assert cache.get() == 5
 ```
 """
 
+from __future__ import annotations
+
 import asyncio
 import typing
 
 from ._receiver import Receiver
 
 T_co = typing.TypeVar("T_co", covariant=True)
+HashableT = typing.TypeVar("HashableT", bound=typing.Hashable)
 
 
 class _Sentinel:
@@ -48,15 +59,59 @@ class _Sentinel:
         return "<no value received yet>"
 
 
-class LatestValueCache(typing.Generic[T_co]):
+_SENTINEL = _Sentinel()
+
+
+class LatestValueCache(typing.Generic[T_co, HashableT]):
     """A cache that stores the latest value in a receiver.
 
     It provides a way to look up the latest value in a stream without any delay,
     as long as there has been one value received.
     """
 
+    @typing.overload
     def __init__(
-        self, receiver: Receiver[T_co], *, unique_id: str | None = None
+        self: LatestValueCache[T_co, _Sentinel],
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: _Sentinel = _SENTINEL,
+    ) -> None:
+        """Create a new cache that does not use keys.
+
+        Args:
+            receiver: The receiver to cache.
+            unique_id: A string to help uniquely identify this instance. If not
+                provided, a unique identifier will be generated from the object's
+                [`id()`][id]. It is used mostly for debugging purposes.
+            key: This parameter is ignored when set to `None`.
+        """
+
+    @typing.overload
+    def __init__(
+        self: LatestValueCache[T_co, HashableT],
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: typing.Callable[[T_co], HashableT],
+    ) -> None:
+        """Create a new cache that uses keys.
+
+        Args:
+            receiver: The receiver to cache.
+            unique_id: A string to help uniquely identify this instance. If not
+                provided, a unique identifier will be generated from the object's
+                [`id()`][id]. It is used mostly for debugging purposes.
+            key: A function that takes a value and returns a key to group the values by.
+                If provided, the cache will store the latest value for each key separately.
+        """
+
+    def __init__(
+        self,
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: typing.Callable[[T_co], typing.Any] | _Sentinel = _SENTINEL,
     ) -> None:
         """Create a new cache.
 
@@ -65,10 +120,16 @@ class LatestValueCache(typing.Generic[T_co]):
             unique_id: A string to help uniquely identify this instance. If not
                 provided, a unique identifier will be generated from the object's
                 [`id()`][id]. It is used mostly for debugging purposes.
+            key: An optional function that takes a value and returns a key to group the
+                values by. If provided, the cache will store the latest value for each
+                key separately. If not provided, it will store only the latest value
+                received overall.
         """
         self._receiver = receiver
+        self._key: typing.Callable[[T_co], HashableT] | _Sentinel = key
         self._unique_id: str = hex(id(self)) if unique_id is None else unique_id
-        self._latest_value: T_co | _Sentinel = _Sentinel()
+        self._latest_value: T_co | _Sentinel = _SENTINEL
+        self._latest_value_by_key: dict[HashableT, T_co] = {}
         self._task = asyncio.create_task(
             self._run(), name=f"LatestValueCache«{self._unique_id}»"
         )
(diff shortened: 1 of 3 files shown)


src/frequenz/channels/_latest_value_cache.py
```diff
@@ -48,15 +57,56 @@ def __str__(self) -> str:
         return "<no value received yet>"
 
 
-class LatestValueCache(typing.Generic[T_co]):
+class LatestValueCache(typing.Generic[T_co, HashableT]):
```
llucax, line 71: Maybe not a biggie, but this introduces a new requirements for values to cache, that will be required even if `key` is not desired or used. I guess in most cases this will not be an issue, and I guess there is no solution unless we have a separate non-keyed `LatestValueCache`, so it is probably fine.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -48,15 +57,56 @@ def __str__(self) -> str:
...
+                If provided, the cache will store the latest value for each key separately.
+        """
+
     def __init__(
-        self, receiver: Receiver[T_co], *, unique_id: str | None = None
+        self,
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: typing.Callable[[T_co], typing.Any] | None = None,
```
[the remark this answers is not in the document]
shsms: No, it has to be a superset of the two overloads `HashableT` and `None`.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -48,15 +57,56 @@ def __str__(self) -> str:
...
+        """
+
+    @typing.overload
+    def __init__(
+        self: "LatestValueCache[T_co, HashableT]",
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: typing.Callable[[T_co], HashableT],
+    ) -> None:
```
llucax: Damn, is this overload over `self` really necessary? Doesn't `self` just works magically if you don't annotate it at all? Have you tried with `Self`?

Also, you should `from __future__ import annotations` and remove the quotes from the types if you still need it.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -78,34 +134,53 @@ def unique_id(self) -> str:
         """The unique identifier of this instance."""
         return self._unique_id
 
-    def get(self) -> T_co:
+    def get(self, key: HashableT | None = None) -> T_co:
```
llucax: So we can't use `None` as a key :thinking: 

I guess eventually we'll want it, so maybe implement this using a sentinel from the start? :D

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -78,34 +134,53 @@ def unique_id(self) -> str:
...
         Returns:
             The latest value that has been received.
 
         Raises:
             ValueError: If no value has been received yet.
         """
+        if key is not None:
+            if key not in self._latest_value_by_key:
+                raise ValueError(f"No value received for key: {key!r}")
+            return self._latest_value_by_key[key]
```
llucax, line 168: If you use a sentinel, you can probably also get rid of the "global" `self._latest_value` and use `self._latest_value_by_key[_Sentinel]` instead as the global value, it could even simplify the code quite a bit as you get rid of the special casing.

src/frequenz/channels/_latest_value_cache.py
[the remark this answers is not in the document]
llucax: But not `HashableT | None`?

src/frequenz/channels/_latest_value_cache.py
shsms: If no key is specified, this will be `None`.  Apparantly `None` is hashable.

src/frequenz/channels/_latest_value_cache.py
shsms: This is how to avoid having to explicitly specify HashableT and default to None when no key is specified.

I will remove the quotes.

llucax: I wonder if using a (non-hashable) sentinel in the type could help with this too, if the issue is that `None` is `Hashable`.

src/frequenz/channels/_latest_value_cache.py
shsms: True, we have a tendency to do weird things.  Done.

src/frequenz/channels/_latest_value_cache.py
shsms: I just tried this and neither the number of if statements nor the number of lines come down.  In addition to that we're having to hash an extra param.  So I don't see the point.

It would have been different, if the place holder in the init wasn't for a key /function/, but rather for a key, I guess, not sure.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -32,31 +40,84 @@
...
     def __str__(self) -> str:
         """Return a string representation of this sentinel."""
-        return "<no value received yet>"
+        return f"<Sentinel: {self._desc}>"
 
 
-class LatestValueCache(typing.Generic[T_co]):
+NO_KEY = Sentinel("no key provided")
+NO_KEY_FUNCTION = Sentinel("no key function provided")
+NO_VALUE_RECEIVED = Sentinel("no value received yet")
```
llucax: Nitpick:
```suggestion
NO_KEY: Final[Sentinel] = Sentinel("no key provided")
NO_KEY_FUNCTION: Final[Sentinel] = Sentinel("no key function provided")
NO_VALUE_RECEIVED: Final[Sentinel] = Sentinel("no value received yet")
```

But nice, it makes it much more readable indeed :tada:

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -32,31 +40,84 @@
...
     as long as there has been one value received.
     """
 
+    @typing.overload
+    def __init__(
+        self: LatestValueCache[T_co, Sentinel],
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: Sentinel = NO_KEY_FUNCTION,
```
llucax, line 84: Can you use `Literal[NO_KEY_FUNCTION]` here instead of `Sentinel`? Otherwise this will also accept `key=NO_VALUE_RECEIVER` for example. I guess it is still OK (the alternative is to create one type per sentinel), but if `Literal` works it could be a very nice solution.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -32,31 +40,84 @@
...
+            key: A function that takes a value and returns a key to group the values by.
+                If provided, the cache will store the latest value for each key separately.
+        """
+
+    def __init__(
+        self,
+        receiver: Receiver[T_co],
+        *,
+        unique_id: str | None = None,
+        key: typing.Callable[[T_co], typing.Any] | Sentinel = NO_KEY_FUNCTION,
```
llucax, line 120: Same here in case `Literal[NO_KEY_FUNCTION]` works.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -65,10 +126,16 @@ def __init__(
...
+            key: An optional function that takes a value and returns a key to group the
+                values by. If provided, the cache will store the latest value for each
+                key separately. If not provided, it will store only the latest value
+                received overall.
         """
         self._receiver = receiver
+        self._key: typing.Callable[[T_co], HashableT] | Sentinel = key
         self._unique_id: str = hex(id(self)) if unique_id is None else unique_id
-        self._latest_value: T_co | _Sentinel = _Sentinel()
+        self._latest_value: T_co | Sentinel = NO_VALUE_RECEIVED
```
llucax, line 137: `Literal[NO_VALUE_RECEIVED]`, etc...

llucax (dismissed):
Some suggestions for extra type-safety, but LGTM.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -78,34 +145,53 @@ def unique_id(self) -> str:
...
+                provided, it retrieves the latest value received overall.
+
         Returns:
             The latest value that has been received.
 
         Raises:
             ValueError: If no value has been received yet.
         """
-        if isinstance(self._latest_value, _Sentinel):
+        if not isinstance(key, Sentinel):
```
llucax, line 165: If `Literal` works, then here (and in other places using `isinstance`) it might be more readable to use `if key is NO_KEY`.

src/frequenz/channels/_latest_value_cache.py
shsms: No, Literal is only available for fundamental types.  I think it is fine, the sentinels are interchangeable and their meaning comes from context, not from their names.  Also, they are not exposed, so users can't use them without accessing private symbols.

src/frequenz/channels/_latest_value_cache.py
shsms: done

src/frequenz/channels/_latest_value_cache.py
shsms: No, same as above.

src/frequenz/channels/_latest_value_cache.py
shsms: Not possible.

src/frequenz/channels/_latest_value_cache.py
shsms: No, doesn't.

llucax (approved):

src/frequenz/channels/_latest_value_cache.py
llucax: And `Enum`s, that could be an alternative, but it is probably not worth it, if the values are not exposed anyway, then it doesn't add much.

PR #426, opened 2025-06-16
Support clearing the latest value in the `LatestValueCache`

shsms:
This can be the overall latest value or the latest value for a
specific key.  This can be used to clean up old keys for which no
further messages are expected.

### DIFF
--- a/src/frequenz/channels/_latest_value_cache.py
+++ b/src/frequenz/channels/_latest_value_cache.py
@@ -194,6 +194,22 @@ class LatestValueCache(typing.Generic[T_co, HashableT]):
             return key in self._latest_value_by_key
         return not isinstance(self._latest_value, Sentinel)
 
+    def clear(self, key: HashableT | Sentinel = NO_KEY) -> None:
+        """Clear the latest value or the latest value for a specific key.
+
+        If `key` is provided, it clears the latest value for that key. If no key is
+        provided, it clears the latest value received overall.
+
+        Args:
+            key: An optional key to clear the latest value for that key. If not
+                provided, it clears the latest value received overall.
+        """
+        if not isinstance(key, Sentinel):
+            _ = self._latest_value_by_key.pop(key, None)
+            return
+
+        self._latest_value = NO_VALUE_RECEIVED
+
     async def _run(self) -> None:
         async for value in self._receiver:
             self._latest_value = value
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -12,6 +12,8 @@
 
 - `LatestValueCache` now takes an optional `key` function, which returns the key for each incoming message, and the latest value for each key is cached and can be retrieved separately.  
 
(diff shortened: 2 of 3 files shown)


shsms:
No release notes, because this is an enhancement to an unreleased feature.

llucax (dismissed):

llucax:
> No release notes, because this is an enhancement to an unreleased feature.

Not entirely true, `LastValueCache` existed already, and this extends the released implementation to allow for clearing the "global" cache.

llucax:
I removed the skip notes label, I'm on the edge here, it is still a pretty small change so not putting it into the release notes is not the end of the world, but technically we should. If you want to add the label again and merge, I won't complain.

shsms:
True, updated.

src/frequenz/channels/_latest_value_cache.py
```diff
@@ -194,6 +194,22 @@ def has_value(self, key: HashableT | Sentinel = NO_KEY) -> bool:
...
+        Args:
+            key: An optional key to clear the latest value for that key. If not
+                provided, it clears the latest value received overall.
+        """
+        if not isinstance(key, Sentinel):
+            _ = self._latest_value_by_key.pop(key, None)
+            return
+
+        self._latest_value = NO_VALUE_RECEIVED
+
```
ela-kotulska-frequenz, line 212: Why call `clear()` doesn't remove all keys and values in `self._latest_value_by_key` ? 
Or shouldn't you remove `_latest_value_by_key[self._key(self._latest_value)]]`, too?

shsms: oh damn, because my usecase didn't need clearing all keys.  I wonder if we should change the API slightly to keep the option open for a "clear everything" in the future.

any opinions @llucax ?

shsms: Another option is to make the `key` parameter required, so that I can make progress and we can decide how to generalise later.

llucax: Some options I can think of:

1. `clear_all()`, `clear()` but it is true that `clear()` with no key looks ambiguous
2. `clear_all()`, `clear_key()` maybe is more clear that without a key it clears the no-key?

I'm starting to find the use case for a "global" key a bit weird. I guess we need to support it for backwards compatibility, otherwise maybe we should always require a key and let the user use something like `None` if they don't want one? Not sure.

shsms: We haven't released it.  I can actually pull out all the new stuff into `experimental.GroupingLatestValueCache` that doesn't have a global mode.

shsms: Done here: https://github.com/frequenz-floss/frequenz-channels-python/pull/428

shsms:
This will go away once https://github.com/frequenz-floss/frequenz-channels-python/pull/428 is merged.

PR #428, opened 2025-06-16
Implement the experimental `GroupingLatestValueCache`

shsms:
It is similar to the `LatestValueCache`, but accepts an additional key-function as an argument, which takes each incoming message and returns a key for that message.  The latest value received for each unique key gets cached and is available to look up on demand.

This also reverts the (unreleased) group-by-key extensions made to `LatestValueCache`, so that there is more time to finalise the API before making it stable.


src/frequenz/channels/experimental/_grouping_latest_value_cache.py
```diff
@@ -0,0 +1,155 @@
...
+        """
+        return key in self._latest_value_by_key
+
+    def clear(self, key: HashableT) -> None:
+        """Clear the latest value or the latest value for a specific key.
+
+        Args:
+            key: The key for which to clear the latest value.
+        """
+        _ = self._latest_value_by_key.pop(key, None)
```
ela-kotulska-frequenz: Probably not udpated documentation: "Clear the latest value for a specific key"

ela-kotulska-frequenz (commented):
LGTM :)

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
shsms: true, fixed.

ela-kotulska-frequenz (approved):

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
```diff
@@ -0,0 +1,155 @@
...
+                provided, it retrieves the latest value received overall.
+
+        Returns:
+            The latest value that has been received.
+
+        Raises:
+            ValueError: If no value has been received yet.
+        """
+        if key not in self._latest_value_by_key:
+            raise ValueError(f"No value received for key: {key!r}")
```
llucax, line 112: Any reason why this is not a `KeyError`? :thinking:

llucax (commented):
I was stating to write a lot of comments about using `__contains__` instead of `has_key`, etc. but then probably we should just inherit from [`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) so we are required to follow the mapping interface, I think it is a perfect fit, this is basically a `dict` that is mutable/filled only externally by the passed receiver.

shsms:
> but then probably we should just inherit from [`Mapping`](https://docs.python.org/3/library/collections.abc.html#collections.abc.Mapping) so we are required to follow the mapping interface

That's a good idea, I'll try.

PR #429, opened 2025-06-18
Use asyncio.Event instead of asyncio.Future in NopReceiver

ela-kotulska-frequenz:
It works the same but is less low-level-ish.


shsms:
I was worried `Event`s might have spurious wakeups like condition variables, but that doesn't appear to be the case, according to the docs.

llucax (approved):

llucax:
> I was worried `Event`s might have spurious wakeups like condition variables, but that doesn't appear to be the case, according to the docs.

I thought they are just a simple wrapper over a future, but apparently they are more complicated because they support multiple waiters.

* https://github.com/python/cpython/blob/5f6ab924653a44e08be23710d5023566e9e9214e/Lib/asyncio/locks.py#L158-L216

PR #430, opened 2025-06-18
Implement the `Mapping` interface for `GroupingLatestValueCache`

shsms:


src/frequenz/channels/experimental/_grouping_latest_value_cache.py
```diff
@@ -39,20 +39,21 @@
...
-from collections.abc import Set
+from collections.abc import Iterator, KeysView, Mapping
+
+from typing_extensions import override
 
 from .._receiver import Receiver
 
 T_co = typing.TypeVar("T_co", covariant=True)
+T = typing.TypeVar("T")
 HashableT = typing.TypeVar("HashableT", bound=typing.Hashable)
```
llucax: Optional: Document these (adding a docstring below them), so they are shown in the docs index.

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
```diff
@@ -39,20 +39,21 @@
...
 
 
-class GroupingLatestValueCache(typing.Generic[T_co, HashableT]):
-    """A cache that stores the latest value in a receiver.
-
-    It provides a way to look up the latest value in a stream without any delay,
-    as long as there has been one value received.
-    """
+class GroupingLatestValueCache(
+    typing.Generic[T_co, HashableT], Mapping[HashableT, T_co]
```
llucax: It shouldn't be necessary to keep the `typing.Generic[T_co, HashableT]` AFAIK, you have them indirectly via `Mapping[HashableT, T_co]`.

llucax (commented):
Not sure if it is worth keeping `c.has_key(k)` instead of requiring the use of `k in c`.

Also not sure if it might be worth implementing also the methods automatically mixed in by `Mapping` and its subclasses: `__contains__`, `items`, `values`, `__eq_`_, and `__ne__`, otherwise they are probably calculated via the iterator or something like that, which could be pretty inefficient.

image
image

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
shsms: They are internal, not exposed from the `_grouping_latest_value_cache` module.  Not sure if they should be though.

shsms:
> Not sure if it is worth keeping `c.has_key(k)` instead of requiring the use of `k in c`.
> 
> Also not sure if it might be worth implementing also the methods automatically mixed in by `Mapping` and its subclasses: `__contains__`, `items`, `values`, `__eq_`_, and `__ne__`, otherwise they are probably calculated via the iterator or something like that, which could be pretty inefficient.

All done.

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
shsms: done

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
llucax: I guess both are fine. If they are not exported, the docs will show it as a "string" without any links, and it will not be present in the ToC. If the types are self explanatory we can keep it as it is. If they are not documented/exported, users will not know about the possible `bound=` arguments or *constraints* (or if it is co/contravariant, but for those we usually use the suffixes), so if it is important to the user to know any of that info, that could be a good reason to export and document.

llucax (dismissed):

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
shsms: I've added a commit to do this.  I'm somewhat doubtful about exposing them from the `experimental` package directly, but that seems to be the best option available.

llucax (approved):

PR #431, opened 2025-06-19
select: Don't try selecting on a closed event loop

Marenz:
With this change, we don't try to clean up any tasks (after the eventloop has ended) as we don't even try to create a task and don't enter the `try/finally` block if there is no event loop at all.


florian-wagner-frequenz (approved):

Marenz:
closing, I don't think it actually helps

PR #433, opened 2025-06-25
Improve `GroupingLatestValueCache`

llucax:
Fixes some documentation issues and brings the interface more close to `MutableMapping`.

### DIFF
--- a/src/frequenz/channels/experimental/_grouping_latest_value_cache.py
+++ b/src/frequenz/channels/experimental/_grouping_latest_value_cache.py
@@ -37,7 +37,8 @@ class GroupingLatestValueCache(Mapping[HashableT, ValueT_co]):
     a [`MutableMapping`][collections.abc.MutableMapping] because users can't mutate the
     cache directly, it is only mutated by the underlying receiver. There is one exception
     though, users can clear individual keys from the cache using the
-    [clear][frequenz.channels.experimental.GroupingLatestValueCache.clear] method.
+    [__delitem__][frequenz.channels.experimental.GroupingLatestValueCache.__delitem__]
+    method.
 
     Example:
         ```python
@@ -56,6 +57,11 @@ class GroupingLatestValueCache(Mapping[HashableT, ValueT_co]):
 
         assert 6 in cache
         assert cache.get(6) == (6, "twenty-six")
+
+        del cache[6]
+
+        assert cache.get(6) is None
+        assert 6 not in cache
 
         await cache.stop()
         ```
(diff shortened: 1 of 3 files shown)


llucax:
Not that I planned to do this, it just happened as I was adapting the code to do make receivers context managers...

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
```diff
@@ -220,6 +231,50 @@ def __delitem__(self, key: HashableT) -> None:
...
+    def pop(self, key: HashableT, /) -> ValueT_co | None:
+        """Remove the latest value for a specific key and return it."""
+
+    @typing.overload
+    def pop(self, key: HashableT, /, default: DefaultT) -> ValueT_co | DefaultT:
+        """Remove the latest value for a specific key and return it."""
+
+    def pop(
+        self, key: HashableT, /, default: DefaultT | _NotSpecified = _NotSpecified()
+    ) -> ValueT_co | DefaultT | None:
```
shsms, line 249:
```suggestion
    def pop(
        self, key: HashableT, /, default: DefaultT | None = None
    ) -> ValueT_co | DefaultT | None:
```

Then you can use `return self._latest_value_by_key.pop(key, default)` as the only body, because that's the exact signature of dict.pop(), and no need for special cases.

And if a user wants to have `None` in `ValueT_co`, they can make their own sentinel that will be `DefaultT`.  So we don't need to make a sentinel to be able to satisfy `MutableMapping`.

llucax: No, the problem is `pop()` raises when the `default` is not specified, it doesn't return `None` (is not symmetric with `get()`), so we need to be able to tell between not passing the argument and passing `None`.

shsms: oh damn, that's such a bad API.

shsms (dismissed):

src/frequenz/channels/experimental/_grouping_latest_value_cache.py
llucax: image

llucax:
Mmm, failing in Python 3.12, will have to have a look.

llucax:
> ************* Module frequenz.channels.experimental._grouping_latest_value_cache
> src/frequenz/channels/experimental/_grouping_latest_value_cache.py:9:0: W4904: Using deprecated class Hashable of module typing (deprecated-class)

Strange that it only fails in 3.12.

Updated with a new commit with a fix.

shsms (approved):

Issue #411, opened 2025-04-22
Improve async robustness and cleanup


llucax:
### What happened?

When we wrote this library, we didn't fully understand the details of `asyncio` and how to do cleanup and cancellation properly. It is not a surprise, as this can be extremely tricky and the documention for `asyncio` is not great at specifying this. Because of this we ended up with some code that either is not cleaning up properly, or it is hard to clean up properly.

### What did you expect instead?

Things to be more robust and clean up properly.

### Affected part(s)

Channels, `Broadcast`, `Anycast`, etc. (part:channels), Core types (`Sender`, `Receiver`, exceptions, etc.) (part:core), Experimental features (the `experimental` package) (part:experimental), Utilities (`Event`, `FileWatcher`, `Timer`, etc.) (part:utilities), Synchronization of multiple sources (`select()`, `merge()`, etc.) (part:synchronization)

### Extra information

We should fix at least these:

- [ ] Remove all `__del__` methods
- [ ] Make all classes that need cleanup a (async) context manager
- [ ] Make `select()` an async context manager (make the function use the context manager and deprecate it)
- [ ] Research if every call to `await` is a *checkpoint*. If not, we should probably add checkpoints to all `async` functions

PR #494, opened 2026-02-09
Automatically closing broadcast channels

shsms:
Channels created with the `broadcast` function will close automatically when either all senders or all receivers to a channel are closed.

### DIFF
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -21,6 +21,43 @@ from ._sender import ClonableSubscribableSender, SenderClosedError, SenderError
 _logger = logging.getLogger(__name__)
 
 
+def broadcast(
+    message_type: type[ChannelMessageT],  # pylint: disable=unused-argument
+    /,
+    *,
+    name: str,
+    resend_latest: bool = False,
+) -> tuple[ClonableSubscribableSender[ChannelMessageT], Receiver[ChannelMessageT]]:
+    """Create a new Broadcast channel and return a sender and a receiver attached to it.
+
+    The channel will be automatically closed when all senders or all receivers
+    are closed.
+
+    Args:
+        message_type: The type of messages that will be sent through this channel. This
+            is only for type checking purposes, it is not used at runtime.
+        name: The name of the channel. This is for logging purposes, and it will be
+            shown in the string representation of the channel.
+        resend_latest: When True, every time a new receiver is created with
+            `new_receiver`, the last message seen by the channel will be sent to the
+            new receiver automatically. This allows new receivers on slow streams to
+            get the latest message as soon as they are created, without having to
+            wait for the next message on the channel to arrive.  It is safe to be
+            set in data/reporting channels, but is not recommended for use in
+            channels that stream control instructions.
+
+    Returns:
+        A tuple of a sender and a receiver attached to the created channel.
+    """
+    channel = Broadcast[ChannelMessageT](
+        name=name, resend_latest=resend_latest, auto_close=True
+    )
+    return channel.new_sender(), channel.new_receiver()
+
+
+@deprecated(
+    "Please use the `broadcast` function to create a Broadcast channel instead."
+)
 class Broadcast(  # pylint: disable=too-many-instance-attributes
     Generic[ChannelMessageT]
 ):
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -21,7 +21,9 @@ from ._sender import ClonableSubscribableSender, SenderClosedError, SenderError
 _logger = logging.getLogger(__name__)
 
 
-class Broadcast(Generic[ChannelMessageT]):
+class Broadcast(  # pylint: disable=too-many-instance-attributes
+    Generic[ChannelMessageT]
+):
     """A channel that deliver all messages to all receivers.
 
     # Description
@@ -184,7 +186,13 @@ class Broadcast(Generic[ChannelMessageT]):
         ```
     """
 
-    def __init__(self, *, name: str, resend_latest: bool = False) -> None:
+    def __init__(
+        self,
+        *,
+        name: str,
+        resend_latest: bool = False,
+        auto_close: bool = False,
+    ) -> None:
         """Initialize this channel.
 
         Args:
@@ -197,6 +205,8 @@ class Broadcast(Generic[ChannelMessageT]):
                 wait for the next message on the channel to arrive.  It is safe to be
                 set in data/reporting channels, but is not recommended for use in
                 channels that stream control instructions.
+            auto_close: If True, the channel will be closed when all senders or all
+                receivers are closed.
         """
         self._name: str = name
         """The name of the broadcast channel.
@@ -220,6 +230,9 @@ class Broadcast(Generic[ChannelMessageT]):
 
         self._latest: ChannelMessageT | None = None
         """The latest message sent to the channel."""
+
+        self._auto_close: bool = auto_close
+        """Whether to close the channel when all senders or all receivers are closed."""
 
         self.resend_latest: bool = resend_latest
         """Whether to resend the latest message to new receivers.
@@ -355,13 +368,20 @@ class _Sender(ClonableSubscribableSender[_T]):
                 set as the cause.
             SenderClosedError: If this sender was closed.
         """
-        if self._closed:
-            raise SenderClosedError(self)
         # pylint: disable=protected-access
         if self._channel._closed:
             raise SenderError("The channel was closed", self) from ChannelClosedError(
                 self._channel
             )
+        if self._channel._auto_close and (
+            self._channel._sender_count == 0 or len(self._channel._receivers) == 0
+        ):
+            await self._channel.aclose()
+            raise SenderError("The channel was closed", self) from ChannelClosedError(
+                self._channel
+            )
+        if self._closed:
+            raise SenderClosedError(self)
         self._channel._latest = message
         stale_refs = []
         for _hash, recv_ref in self._channel._receivers.items():
@@ -507,6 +527,11 @@ class _Receiver(Receiver[_T]):
         # pylint: disable=protected-access
         while len(self._q) == 0:
             if self._channel._closed or self._closed:
+                return False
+            if self._channel._auto_close and (
+                self._channel._sender_count == 0 or len(self._channel._receivers) == 0
+            ):
+                await self._channel.aclose()
                 return False
             async with self._channel._recv_cv:
                 await self._channel._recv_cv.wait()
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -16,7 +16,7 @@ from typing_extensions import deprecated, override
 from ._exceptions import ChannelClosedError
 from ._generic import ChannelMessageT
 from ._receiver import Receiver, ReceiverStoppedError
-from ._sender import Sender, SenderError
+from ._sender import ClonableSubscribableSender, SenderError
 
 _logger = logging.getLogger(__name__)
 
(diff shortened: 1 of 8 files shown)


src/frequenz/channels/_sender.py
```diff
@@ -88,3 +100,47 @@ def __init__(self, message: str, sender: Sender[SenderMessageT_co]):
...
+            sender: The [Sender][frequenz.channels.Sender] that was closed.
+        """
+        super().__init__("Sender is closed", sender)
+
+
+class SubscribableSender(Sender[SenderMessageT_contra], ABC):
+    """A [Sender][frequenz.channels.Sender] that can be subscribed to."""
+
+    @abstractmethod
+    def subscribe(self) -> Receiver[SenderMessageT_contra]:
```
llucax, line 121: Did you considered using `new_receiver()` instead, as it is the name we use everywhere else?

src/frequenz/channels/_sender.py
```diff
@@ -88,3 +100,47 @@ def __init__(self, message: str, sender: Sender[SenderMessageT_co]):
...
+    def __init__(self, sender: Sender[SenderMessageT_co]):
+        """Initialize this error.
+
+        Args:
+            sender: The [Sender][frequenz.channels.Sender] that was closed.
+        """
+        super().__init__("Sender is closed", sender)
+
+
+class SubscribableSender(Sender[SenderMessageT_contra], ABC):
```
llucax, line 117: Nitpick: You don't need to pass `ABC` here, `Sender` is already an `ABC`, so as long as you don't implement its abstract methods, it will still be an abstract class. Not sure if you prefer to mark it explicitly in case something changes in the future.

(same with the other new abstract classes)

src/frequenz/channels/_sender.py
```diff
@@ -70,6 +73,15 @@ async def send(self, message: SenderMessageT_contra, /) -> None:
...
         """
 
+    @abstractmethod
+    def close(self) -> None:
+        """Close this sender.
+
+        After a sender is closed, it can no longer be used to send messages. Any
+        attempt to send a message through a closed sender will raise a
+        [SenderError][frequenz.channels.SenderError].
+        """
```
llucax, line 83: When we talk about closing in an async context, we need to start thinking about asynchronicity in closing too. If we keep close sync (so no `aclose()`), we should probably add a way to wait for completion (for example, flushing a sending queue), but we can add that in the future, when/if needed.

I'm really not sure what is best here. My feeling is splitting *close initiation* (sync) from *close completion* (async) might be more error prone, because people might miss the awaiting for completion, which could potentially end up in wrong finalization or thinking something finished immediately when it actually kept working in the background somehow.

Using `aclose()` makes closing more atomic, and more clear when there is actually a need to wait for something to finish before the close is really complete. But it is less flexible, as one can only close something in an async context, which sometimes can complicate things. But my feeling is this happens because `asyncio` kind of encourages this kind of pattern, they usually split closing from awaiting (like task cancel/await). When using a sender/receiver you should be in an async context anyway.

BTW, Trio has an abstract base class [`AsyncResource`](https://trio.readthedocs.io/en/stable/reference-io.html#trio.abc.AsyncResource) with `aclose()` and everything that is async uses `aclose()`, so if we are in doubt, I would go for `aclose()`, as it seems like the safest approach.

src/frequenz/channels/_broadcast.py
```diff
@@ -207,6 +254,9 @@ def __init__(self, *, name: str, resend_latest: bool = False) -> None:
         self._recv_cv: Condition = Condition()
         """The condition to wait for data in the channel's buffer."""
 
+        self._sender_count: int = 0
```
llucax, line 257: Nitpick: I think we should expose this as a read-only property, it might be useful if someone need to make a more manual tracking of a channel's usage.

src/frequenz/channels/_broadcast.py
```diff
@@ -218,6 +268,9 @@ def __init__(self, *, name: str, resend_latest: bool = False) -> None:
         self._latest: ChannelMessageT | None = None
         """The latest message sent to the channel."""
 
+        self._auto_close: bool = auto_close
```
llucax, line 271: Same, having a way to query if a channel is auto-close or not might be useful. Even more, I wonder if this one could even be read-write. Although probably is not very useful, even if possible, but at least I would expose it as a read-only property.

llucax, line 271: `_auto_close` sounds like a verb, flags should read as a question. I suggest something like `_auto_close_enabled`. It seems silly, but I bumped with this actual issue when suggesting to move the logic to auto close the channel to a separate method, which of course should be called `_auto_close()` 😆

src/frequenz/channels/_broadcast.py
```diff
@@ -345,12 +403,22 @@ async def send(self, message: _T, /) -> None:
...
         """
         # pylint: disable=protected-access
         if self._channel._closed:
             raise SenderError("The channel was closed", self) from ChannelClosedError(
                 self._channel
             )
+        if self._channel._auto_close and (
+            self._channel._sender_count == 0 or len(self._channel._receivers) == 0
+        ):
+            await self._channel.aclose()
```
llucax, line 416: Nitpick: This is used a couple of times, to prevent bugs it might be worth putting in its own method `_auto_close()` (we need to rename the attribute).

src/frequenz/channels/_broadcast.py
```diff
@@ -16,12 +16,51 @@
...
+    """
+    channel = Broadcast[ChannelMessageT](
+        name=name, resend_latest=resend_latest, auto_close=True
+    )
+    return channel.new_sender(), channel.new_receiver()
+
+
+@deprecated(
+    "Please use the `broadcast` function to create a Broadcast channel instead."
+)
```
llucax, line 60: I would skip deprecation for now, let's give the new approach some time to sink in before spamming basically every downstream user. I would even consider adding the new function function to `experimental`, but I guess it still OK to put in the main namespace (if we agree on the name 😆), so I'm fine with that too.

src/frequenz/channels/_broadcast.py
```diff
@@ -16,12 +16,51 @@
...
 from ._generic import ChannelMessageT
 from ._receiver import Receiver, ReceiverStoppedError
-from ._sender import Sender, SenderError
+from ._sender import ClonableSubscribableSender, SenderClosedError, SenderError
 
 _logger = logging.getLogger(__name__)
 
 
-class Broadcast(Generic[ChannelMessageT]):
+def broadcast(
```
llucax, line 24: What about `create_broadcast()` or `make_broadcast()` (we could use `new_broadcast()` to match the current naming convention too, but I think we should use a verb instead). For me it reads weird that `broadcast()` returns a (sender, receiver). Trio actually uses *open* [`open_memory_channel`](https://trio.readthedocs.io/en/stable/reference-core.html#trio.open_memory_channel), which for something that can be closed it doesn't sound like a bad idea: `open_broadcast_channel()`

 Another good thing Trio does, is make senders and receivers async context managers, so they can be used with `async with`. Not for this PR, but we should do that too (#411).

I know I might sound like a broken record citing Trio, but it really looks like they put a **lot** of thought to the design, this is why I think it is a good idea to use it as inspiration. I recommend in particular their [example on using channels](https://trio.readthedocs.io/en/stable/reference-core.html#clean-shutdown-with-channels), as it is very idiomatic for Python, while copying Rust interfaces might not map as well.

tests/test_broadcast.py
```diff
@@ -425,3 +426,50 @@ async def test_broadcast_close_receiver() -> None:
 
     with pytest.raises(ReceiverStoppedError):
         _ = await receiver_2.receive()
+
+
+async def test_broadcast_auto_close_1() -> None:
```
llucax, line 431: Nitpick:
```suggestion
async def test_broadcast_auto_close_all_receivers_closed() -> None:
```

tests/test_broadcast.py
```diff
@@ -425,3 +426,50 @@ async def test_broadcast_close_receiver() -> None:
...
+    assert (await receiver_2.receive()) == 2
+
+    receiver_2.close()
+
+    with pytest.raises(SenderError) as excinfo:
+        await sender.send(3)
+    assert isinstance(excinfo.value.__cause__, ChannelClosedError)
+
+
+async def test_broadcast_auto_close_2() -> None:
```
llucax, line 455: Nitpick:
```suggestion
async def test_broadcast_auto_close_all_senders_closed() -> None:
```

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,90 @@
...
+
+from ._generic import ChannelMessageT
+from ._receiver import Receiver, ReceiverStoppedError
+from ._sender import Sender, SenderClosedError
+
+
+def oneshot(
+    message_type: type[ChannelMessageT],  # pylint: disable=unused-argument
+) -> tuple[Sender[ChannelMessageT], Receiver[ChannelMessageT]]:
+    """Create a one-shot channel.
```
llucax, line 17: I would name this `open_oneshot_channel()` for the same reasons as the broadcast channel.

src/frequenz/channels/_one_shot.py
```diff
@@ -0,0 +1,90 @@
...
+        self._channel.sent = True
+
+
+class _OneShotReceiver(Receiver[ChannelMessageT]):
+    def __init__(self, channel: _OneShot[ChannelMessageT]) -> None:
+        self._channel = channel
+
+    async def ready(self) -> bool:
+        while not self._channel.sent:
+            await self._channel.condition.wait()
```
llucax: This is missing the [`release()`](https://docs.python.org/3/library/asyncio-sync.html#asyncio.Condition):
```suggestion
            async self._channel.condition:
                with await self._channel.condition.wait()
```

src/frequenz/channels/_one_shot.py
```diff
@@ -0,0 +1,90 @@
...
+class _OneShotSender(Sender[ChannelMessageT]):
+    def __init__(self, channel: _OneShot[ChannelMessageT]) -> None:
+        self._channel = channel
+
+    async def send(self, message: ChannelMessageT, /) -> None:
+        if self._channel.sent:
+            raise SenderClosedError(self)
+        self._channel.message = message
+        self._channel.sent = True
+        if self._channel.condition.locked():
```
llucax: Do you really need to check for `locked()`? Can't you always `notify()`?

src/frequenz/channels/_one_shot.py
```diff
@@ -0,0 +1,90 @@
...
+    message is sent, the sender is closed and any further attempts to send a message
+    will raise a `SenderClosedError`.
+    """
+
+    def __init__(self) -> None:
+        """Create a new one-shot channel."""
+        self.message: ChannelMessageT | _Empty = _EMPTY
+        self.sent = False
+        self.drained = False
+        self.condition = Condition()
```
llucax: It looks to me that this could be implemented more simply using an `Event`. Just wait for the event on the receiver and set the event from the sender. Do you really need a `Condition` here?

llucax (commented):
LGTM in general, but there are a few design issues we need to resolve/agree on. I would say the most important points are:

1. `close()` vs `aclose()` (I would really go with `aclose()` unless we have a very good reason not to).
2. Naming.

The PR is also **missing documentation updates**. The docs still shows `Broadcast()` as the only way to create a broadcast channel.

src/frequenz/channels/_broadcast.py
ela-kotulska-frequenz, line 450: If sender is closed and user clones it, then it will create new open (self._closed = False)  sender. Is it ok?
It would mean, that you can create open sender to closed channel.

src/frequenz/channels/_broadcast.py
```diff
@@ -380,6 +385,7 @@ def close(self) -> None:
         [SenderError][frequenz.channels.SenderError].
         """
         self._closed = True
+        self._channel._sender_count -= 1
```
ela-kotulska-frequenz, line 445: What if user doesn't close sender before removing it?
Would it work if we double check if sender is closed in `__del__` method and print warning if not?

llucax:
We can close this one now, right?

shsms:
Yes

PR #502, opened 2026-03-05
Add the oneshot channel implementation

shsms:


src/frequenz/channels/_anycast.py
```diff
@@ -367,6 +374,16 @@ async def send(self, message: _T, /) -> None:
...
 
+    @override
+    async def aclose(self) -> None:
+        """Close this sender.
+
+        After closing, the sender will not be able to send any more messages. Any
+        attempt to send a message through a closed sender will raise a
+        [SenderError][frequenz.channels.SenderError].
+        """
+        self._closed = True
```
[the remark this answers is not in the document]
shsms: Anycast will be deprecated.

src/frequenz/channels/_sender.py
```diff
@@ -70,6 +70,15 @@ async def send(self, message: SenderMessageT_contra, /) -> None:
             SenderError: If there was an error sending the message.
         """
 
+    @abstractmethod
+    async def aclose(self) -> None:
+        """Close this sender.
+
+        After a sender is closed, it can no longer be used to send messages. Any
+        attempt to send a message through a closed sender will raise a
+        [SenderError][frequenz.channels.SenderError].
```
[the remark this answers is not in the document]
shsms: done

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,97 @@
...
+            raise SenderClosedError(self)
+        self._channel.message = message
+        self._channel.sent = True
+        self._channel.event.set()
+
+    async def aclose(self) -> None:
+        self._channel.sent = True
+        if isinstance(self._channel.message, _Empty):
+            self._channel.drained = True
+        self._channel.event.set()
```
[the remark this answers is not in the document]
shsms: fixed.

tests/test_oneshot.py
```diff
@@ -0,0 +1,83 @@
...
+
+
+async def test_oneshot_recv_before_sender_closed() -> None:
+    """Test that closing sender works without sending a message.
+
+    `receiver.receive()` is called before `sender.aclose()`.
+    """
+    sender, receiver = make_oneshot(int)
+
+    task = asyncio.create_task(receiver.receive())
```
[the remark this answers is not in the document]
shsms: done

src/frequenz/channels/_broadcast.py
```diff
@@ -365,6 +371,16 @@ async def send(self, message: _T, /) -> None:
             self._channel._recv_cv.notify_all()
         # pylint: enable=protected-access
 
+    @override
+    async def aclose(self) -> None:
+        """Close this sender.
+
+        After a sender is closed, it can no longer be used to send messages. Any
+        attempt to send a message through a closed sender will raise a
+        [SenderError][frequenz.channels.SenderError].
```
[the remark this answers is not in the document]
shsms: done.

src/frequenz/channels/_sender.py
```diff
@@ -88,3 +97,15 @@ def __init__(self, message: str, sender: Sender[SenderMessageT_co]):
         super().__init__(message)
         self.sender: Sender[SenderMessageT_co] = sender
         """The sender where the error happened."""
+
+
+class SenderClosedError(SenderError[SenderMessageT_co]):
+    """An error indicating that a send operation was attempted a closed sender."""
```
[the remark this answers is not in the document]
shsms: fixed.

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,97 @@
...
+        if self._channel.drained:
+            raise ReceiverStoppedError(self)
+
+        assert not isinstance(
+            self._channel.message, _Empty
+        ), "`consume()` must be preceded by a call to `ready()`."
+
+        self._channel.drained = True
+        self._channel.event.clear()
+        return self._channel.message
```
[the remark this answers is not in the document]
shsms: fixed.

RELEASE_NOTES.md
```diff
@@ -11,7 +11,7 @@
 
 ## New Features
 
-<!-- Here goes the main new features and examples or instructions on how to use them -->
+- There's a new `oneshot` channel, which can be created with the `make_oneshot` function, which returns a sender and a receiver.  A single message can be sent using the sender, after which it will be closed.  And the receiver will close as soon as the message is received.
```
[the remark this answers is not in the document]
shsms: fixed.

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,115 @@
...
+
+    A one-shot channel is a channel that can only send one message. After the first
+    message is sent, the sender is closed and any further attempts to send a message
+    will raise a `SenderClosedError`.
+    """
+
+    def __init__(self) -> None:
+        """Create a new one-shot channel."""
+        self.message: ChannelMessageT | _Empty = _EMPTY
+        self.sent = False
```
simonvoelcker: Feels like `self.sent` could be a derived state of `self.message` instead? Something like:

```python
@property
def sent(self):
  return self.message is not _EMPTY
```

That would make the illegal state of these not being in agreement impossible by design.

llucax: Or `self._channel.event.is_set()`.

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,115 @@
...
+
+import asyncio
+import typing
+
+from ._generic import ChannelMessageT
+from ._receiver import Receiver, ReceiverStoppedError
+from ._sender import Sender, SenderClosedError
+
+
+def make_oneshot(
```
llucax: `open_oneshot_channel()`? 😬 

I feel pretty strongly about this one. This is not a Trio-only thing, open is also used in `asyncio`, like [`open_connection()`](https://docs.python.org/3/library/asyncio-stream.html#asyncio.open_connection).

I would also be up for `open_channel()` and take the type as an argument somehow, but I guess it will make the interface more complicated (it will need more type parameters) with very little gain.

RELEASE_NOTES.md
```diff
@@ -11,7 +11,9 @@
...
+
+- `Sender`s now have an `aclose`, which can be used to close them.
```
```suggestion
- `Sender`s now have an `aclose`, which must be used to close them.
```

😬

src/frequenz/channels/_oneshot.py
```diff
@@ -0,0 +1,115 @@
...
+        """Check if a message is ready to be received.
+
+        Returns:
+            `True` if a message is ready to be received, `False` if the sender
+                is closed and no message will be sent.
+        """
+        if self._channel.drained:
+            return False
+        while not self._channel.sent:
+            await self._channel.event.wait()
```
llucax, line 79: I think the `while` is not really needed here, I guess `await` can only return even if it is not done waiting only if it is cancelled (or other exceptions are raised), but in that case the `while` doesn't matter either. But I guess it doesn't hurt to leave it either.

<details>
<summary>Chat with AI</summary>

For **`asyncio.Event` specifically**, the loop is *not* necessary for “spurious wakeups” the way it is for POSIX condition variables.

`asyncio.Event.wait()` is specified to:

* return immediately if the internal flag is already set
* otherwise block “until another task calls `set()`” ([[Python documentation](https://docs.python.org/3/library/asyncio-sync.html?utm_source=chatgpt.com)][1])

So under normal operation, `await event.wait()` should not resume *without* the event being set. There isn’t an analogue of “POSIX signals interrupting `pthread_cond_wait` and returning success anyway” here: signals may wake the *event loop* at the selector level, but the task awaiting `Event.wait()` won’t complete unless the event gets set (or the task is cancelled / an exception bubbles). If you get a `KeyboardInterrupt`/`CancelledError`, that’s an exception path, not a silent early return.

That means this:

```py
if not self._channel.sent:
    await self._channel.event.wait()
```

is enough **given the current semantics**.

So why do people still write `while predicate: await wait()`?

* It’s a **generic “wait for predicate” pattern** that remains correct if the implementation later changes to a `Condition`, a multi-producer scenario, timeouts, or if someone adds extra `event.set()` calls not strictly tied to `sent`.
* It defends against *logic-level* spurious wakeups (i.e., “event was set for some other reason”) rather than OS-level spurious wakeups.

In your current oneshot code, the event appears to be set only when `sent` transitions true (or `aclose()` forces it), so the loop is mostly **defensive style**, not required for correctness with `asyncio.Event`.

</details>

src/frequenz/channels/_oneshot.py
llucax: Maybe we can get rid of the redundancy (and possibility of ending with inconsistent state) merging `sent`, `event` and `message` into a [`asyncio.Queue(maxsize=1)`](https://docs.python.org/3/library/asyncio-queue.html#queue).

shsms: we also use `sent` to track if a channel was closed before sending anything.  I guess I should change to use a different field for tracking closed, then.

RELEASE_NOTES.md
simonvoelcker: ... unless the sender was closed automatically by sending a message through it, I guess? So **can** would be the correct wording?

simonvoelcker (dismissed):

src/frequenz/channels/_oneshot.py
shsms: I'm actually inclined to switching back to `channels.oneshot` and `channels.broadcast`.

IMHO, `open` doesn't work for us at all.  For files and networking, open is an idiomatic verb.  In both those cases, there is a counterparty or a resource owner, and the idea is to open something to access the other side.

What we are doing is more like creating, because we are creating both sides and the link between them.  So to me, Trio's naming is imprecise.

I'd take inspiration from [os.pipe](https://docs.python.org/3/library/os.html#os.pipe), which is so much closer to what we're doing.

shsms: That would also match Rust, but I'm ok with `make_oneshot`, etc. as well, and it also somewhat similar to the Go syntax, where you have to call `make` for many things, including channels, a bit like the `new` keyword of C++.

RELEASE_NOTES.md
shsms: I think we want users to close things that are not needed anymore, even though we will take some precaution to cleanup.  Because python lacks deterministic or async destructors, cleanup must be done by the users.

I will update the note.

src/frequenz/channels/_oneshot.py
shsms: I didn't make any logic changes, just renamed the variable from `sent` to `closed`, and now I think it reads much better.

RELEASE_NOTES.md
shsms: Done, also rephrased a little.

src/frequenz/channels/_oneshot.py
shsms: I guess it is fine to be conservative here, because this is used mostly for setup and won't be part of any tight loop.

src/frequenz/channels/_oneshot.py
llucax: `oneshot` for me is not an option because it is not a verb. That leaves us with make or open. Let's say we forget about trio and asyncio, I don't understand why you want to introduce an asymmetry between make/close instead of going with open.

And you are opening a channel, the fact that you hide the channel object l, and you close it indirectly by closing all it's sender's or all it's receivers doesn't change the fact that you are opening a channel that needs to be closed.

And I don't get why you want to follow other programming languages naming conventions in python. I really think is not very relevant what rust or go does here.

I think `os.pipe` is not a good example because it follows very old POSIX naming conventions, not python ones.

I think `open_oneshot_channel` is the most precise, descriptive and pythonic option.

src/frequenz/channels/_oneshot.py
llucax: It definitely reads much better with closed. I guess with this all 3 "states" are mostly independent (it can be closed at any time, and the event can be set and drained or not drained, I guess the only dependency is drained can't be true if the event is not set, but that's just a very small dependency).

src/frequenz/channels/_oneshot.py
shsms: > I don't understand why you want to introduce an asymmetry between make/close instead of going with open.

Only because we'd be sacrificing accuracy in favour of symmetry, which doesn't sound like a good idea.  We don't have to open a thing to close it.  We just need the thing to be open so that we can close it.  And that's what the method would do.  

When you create a channel, you get the two open ends of it, that you can close, because when they are created, they are in the open state.

To "open" a thing, its previous state needs to be "closed".  To "make" or "create" a thing, it shouldn't exist previously.

llucax: OK, I was going to go with another round of defending "open" with mostly the same arguments, but instead I decided to ask AI to fact-check this to see which arguments are really based on real evidence.

> The strongest **existing** argument in the thread is Sahas’s objection to using **symmetry as the primary reason**. “It matches `close`” is nice, but by itself it is not enough to justify a public API name. That part of the rebuttal is solid.
> 
> Where I think the discussion becomes stronger is this distinction:
> 
> * `asyncio.open_connection()` is about **establishing a connection to an already-existing external endpoint** and returning endpoint objects. ([Python documentation][2])
> * But factories that **create both local endpoints** in Python are often named like `os.pipe()`, `socket.socketpair()`, or `multiprocessing.Pipe()`, all of which create and return both ends directly, without using `open`. ([Python documentation][3])
> * On the other hand, Trio is the closest conceptual precedent for a channel library, and Trio explicitly uses `open_memory_channel()` for an in-memory channel that returns `(send_channel, receive_channel)` and is later closed. ([trio.readthedocs.io][4])
> 
> So I would separate the question into two parts:
> 
> 1. **Is `open` inaccurate?**
>    No. Trio shows it is perfectly defensible for channels. ([trio.readthedocs.io][4])
> 
> 2. **Is `open` the only or clearly best accurate verb?**
>    Also no. Python has strong counterexamples for paired-endpoint factories that use `pipe`/`socketpair`/`Pipe` rather than `open`. ([Python documentation][3])
> 
> ## Which side is strongest?
> 
> If I had to pick the strongest **substantive** argument, it is this one:
> 
> > The function creates a fresh in-memory communication primitive and returns both ends, so “create/make” has a legitimate semantic claim, and “symmetry with close” is not enough to overrule that.
> 
> That is stronger than the symmetry argument alone. The part I would **not** rely on is the strict semantic rule “to open, the thing must previously have been closed”; real APIs do not follow that rule consistently, and Trio directly contradicts it for channels.
> 
> ## How I would resolve it
> 
> I would name it:
> 
> 
> `open_oneshot_channel()`
> 
> Reason:
> 
> * This is a **channel library**, so Trio’s `open_memory_channel()` is the closest precedent, closer than `os.pipe()` or `socket.socketpair()`. ([trio.readthedocs.io][4])
> * It is clearer than `make_oneshot()`, which omits “channel” and sounds a bit Go-ish rather than Pythonic. The current PR text itself describes it as creating a “one-shot channel,” which suggests “channel” belongs in the public name. ([code][5])
> * It is more descriptive than bare `oneshot`, which is not a verb and reads poorly as a factory name. That objection in the thread is correct.
> 
> If `open_*` is rejected, my fallback would **not** be `make_oneshot()`. I would use:
> 
> `create_oneshot_channel()`
> 
> That would be semantically cleaner than `make_oneshot()`.

[2]: https://docs.python.org/3/library/asyncio-stream.html "Streams — Python 3.14.3 documentation"
[3]: https://docs.python.org/3/library/os.html "os — Miscellaneous operating system interfaces — Python 3.14.3 documentation"
[4]: https://trio.readthedocs.io/en/stable/reference-core.html "Trio’s core functionality — Trio 0.32.0 documentation"
[5]: https://raw.githubusercontent.com/shsms/frequenz-channels-python/oneshot/src/frequenz/channels/_oneshot.py "raw.githubusercontent.com"

(beware, there is scrolling in the quoted text above)

For me the most compelling argument against open is my main reference is `open_connection()` is actually inaccurate (but AI was better explaining why :laughing:)

> * `asyncio.open_connection()` is about **establishing a connection to an already-existing external endpoint** and returning endpoint objects. ([Python documentation][2])
> * But factories that **create both local endpoints** in Python are often named like `os.pipe()`, `socket.socketpair()`, or `multiprocessing.Pipe()`, all of which create and return both ends directly, without using `open`. ([Python documentation][3])
> * On the other hand, Trio is the closest conceptual precedent for a channel library, and Trio explicitly uses `open_memory_channel()` for an in-memory channel that returns `(send_channel, receive_channel)` and is later closed. ([trio.readthedocs.io][4])

So, if we ignore Trio, which I think is a valid approach (although again, being a much better and well thought library than the built-in asyncio, I really think you should read the docs before dismissing it, I think it would also make sense to follow it instead of ignoring it), the argument for *open* is weak. I see that now.

So I feel much better about dropping "open" now, but I also feel more strong against "make_oneshot" :laughing:. The reasons against it in the AI response (paraphrasing it):

> * `make_oneshot()` omits “channel” and sounds a bit Go-ish rather than Pythonic. The current PR text itself describes it as creating a “one-shot channel,” which suggests “channel” belongs in the public name. ([code][5])
> * It is more descriptive than bare `oneshot`, which is not a verb and reads poorly as a factory name. That objection in the thread is correct.
> 
> If `open_*` is rejected, my fallback would **not** be `make_oneshot()`. I would use:
> 
> `create_oneshot_channel()`
> 
> That would be semantically cleaner than `make_oneshot()`.

I really think we should include `_channel` and we should use something else other than `make_`. I think `create_` is a solid alternative, and we already use it in some other places (like the [SDK](https://github.com/search?q=repo%3Afrequenz-floss%2Ffrequenz-sdk-python+create&type=code&p=2), although for internal stuff mostly). Another alternative we use very widely is `new_`, but we want to move away from it as it is not a verb.

But if this is about negotiation and I have to give on `make_` to get `_channel`, I guess I can take that hit, but it really feels too Go-ish, very out of place in Python IMHO. :P

shsms: I'm completely ok with `create_*`.  

But I'm skeptical about the suffix, because this is part of the channels package.  `channels.create_oneshot` sounds great to me, and here our situation is different from trio, because they can't have `trio.open_memory`, because that would mean something completely different.  So they really need it.  For us it is extra info that makes the name more clunky than necessary.  What does AI say about the suffix.

shsms: Although, I'm not sure I'd trust AI's intuitions in this case either.  I feel less strongly about this than about `open_*`, so I can be persuaded to change my mind, but I'd say again that our situation is not comparable to trio in this case.

shsms: By that logic, we'll have to add suffixes to timer, filewatcher, etc as well.  I think it is fine without the suffix, because of the library's name and people can look at the docs if they don't understand the name.

shsms: Actually `create_oneshot` doesn't work so well as `make_oneshot`, which I think is fine.  Go can't claim ownership over `make`, it is a regular word and it fits us great.

So my preferences in order are -

1. `make_oneshot`
2. `make_oneshot_channel`
3. `create_oneshot_channel`
4. `create_oneshot`

llucax: > But I'm skeptical about the suffix, because this is part of the channels package. `channels.create_oneshot` sounds great to me

I agree in concept, but the problem is the style of using packages didn't stick, and we are almost never using `import pkg; pkg.sym` and instead we are using almost always `from pkg import sym; sym`. I think we have using the first form somewhere in the coding style but we never really used it. I had a few attempts to start using it but again, it didn't stick. I think `import pkg` is widely used in Rust and Go (if not the only way to do it), but I think using `from pkg import sym` is stick just too deep in python devs and this will be too hard to change, if not impossible, so I would give up on this one.

But see also: https://github.com/frequenz-floss/frequenz-sdk-python/pull/1371#discussion_r2905893867. It seems there are more issues with this interface. I really wonder if the `NamedTuple` trick works. Then it would just be `OneshotChannel`.

llucax: > By that logic, we'll have to add suffixes to timer, filewatcher, etc as well. I think it is fine without the suffix, because of the library's name and people can look at the docs if they don't understand the name.

I think the difference here is "one shot" is an adjective because you don't mean literally creating one "shot", one shot of what? You are creating a "one-shot channel". FileWatcher and Timer are nouns. Same with broadcast, the noun would be "broadcaster" but that also sounds more like an actor that broadcasts things, not the medium used to send the broadcast.

Also, for receivers is almost the reverse, the fact that is a receiver is just like and adjective, it means it follows an interface, but what the object does is actually tracking time or watching files, not "receiving". With channels is the opposite, there is no "channel interface", being a channel is the main thing, and the one shot or broadcast is just a "detail" in a way.

llucax: About `make` vs `create`, I think you are just trolling now :laughing: .

Create is much more widely used. We don't use `make` anywhere that I'm aware of, and there is no built-in native[^1] python function using it, while there are 42 `create_*()` (including many in `asyncio` (asyncio.create_task(), asyncio.create_subprocess_exec(), asyncio.create_subprocess_shell(), loop.create_task(), loop.create_connection(), loop.create_datagram_endpoint(), loop.create_server(), loop.create_unix_connection(), loop.create_unix_server(), loop.create_future()).

But again, maybe we can get rid of this problem if we change the interface to create a namedtuple instead and we just call it `OneshotChannel`.

And if I get more dramatic, considering the main use for this was to replace the channel registry and I think that is not really working great, I'm not sure we really need a oneshot channel at all :scream: 

[^1]: As in not inherited from gettex or TK and even counting those there are only a couple.

shsms:
The idea to try with a NamedTuple was a good hint.  Didn't need NamedTuple, just derived from `tuple` and it looks great.  Called it `Oneshot`, to be similar to `Broadcast`.

Will update `Broadcast` to support this pattern also.

EDIT: maybe will rename to `OneshotChannel` and `BroadcastChannel`, so that we don't break the `Broadcast` interface.  Or we discuss other options.

src/frequenz/channels/_oneshot.py
shsms: >  I think that is not really working great

I haven't read your issues in detail yet.  Maybe talk about it tomorrow morning?

simonvoelcker: And I though **I** was serious about naming :D

Maybe I can act as a tie breaker for you. I think that:
- `<verb>_oneshot_channel` is the most pythonic pattern. I've noticed that short names, and even abbreviations (e.g. "recv" for receiver) are widely used at Frequenz, which I haven't seen much in past jobs. 
- IMO other languages do not matter unless we plan to migrate this to Rust soon and find it helpful to use Rust's naming conventions already.

And just thinking out loud, have you considered something like this?

```python
sender, receiver = Oneshot.create_sender_and_receiver()
```

llucax:
OK, I have a lot of mixed feelings and now looking at it implemented, for me the interface still looks a bit weird (instantiating a channel giving you 2 things), but I can definitely live with it. To try to move forward, I will make 2 small comments:

* I think adding the `Channel` suffix is a good addition (because of the `from channels import ...` thing, so I hope you do that.
* A maybe, just maybe, way to make it less weird could be `sender, receiver = OneshotChannelPair[int]()` (not a lot of existing examples, but there is already a `sockerpair()` out there).

EDIT: Or just `OneshotPair`, as it is not really a pair of channels, so I can live without the channel word in there. But 

<details>
<summary>Comment that doesn't help to keep progressing, but for me makes a very good argument :laughing: </summary>

I don't know anymore really. I still think the `open_oneshot_channel` is the cleanest, because the action is opening a channel, and the result is the creation of a sender and a receiver pair. "create" for me is too literal, "create_channel" sounds like you should get a channel, not a sender and receiver, "create_sender_and_receiver" puts too much weight into the sender and receiver, you are creating a channel here, is not like you can have a "create_sender" or "create_receiver" only, they are both ends of a channel.

</details>

src/frequenz/channels/_oneshot.py
llucax: That would mean we need to make the channel class public, but we want to keep it private. But as you might have guessed, I like the explicitness. Maybe to make it a bit shorter it could be `create_pair()`, and then there is my suggestion above to go with `OneshotChannelPair`.

llucax: > And I though **I** was serious about naming :D

Yeah, I have mixed feelings about it, for one side it feels irrelevant, for the other hand is a bit of time invested to get a clean and clear API, which is hard to change in the future.

shsms:
Awesome, no you were right that having to pass the type as an argument is a big disadvantage with using a function.  And I'm really glad that subclassing a tuple worked.  We can call them `OneshotChannel` and `BroadcastChannel`.  I'll just add an example and docs and make the PR ready.

llucax:
OK, this looks good to go to me, but I really wonder if it is the best tool to fix the channel registry issue, and if we don't need it for that, I guess we don't have a use case for it, and if we are talking about deprecating/removing anycast because we don't have a use case for it, I'm a bit hesitant to add new channels types we don't use, so I would hold the merging until we are sure we need it. If we decide it is not the right solution, we can merge it anyway in `experimental`, this way it is there in case we find a good use for it, but it doesn't get promoted as a main feature of the lib.

llucax (approved):
OK, we agreed to move forward with this, so approving. But I want to leave it on record that I still have some serious doubts this is a good move.

Time will tell. 🤷

PR #503, opened 2026-03-10
Auto closing broadcast

shsms:
## Pull request overview

This PR introduces a new auto-closing broadcast channel API (`BroadcastChannel`) and deprecates the legacy `Broadcast`/`Anycast` channel classes, aiming to make broadcast channels close automatically when one side (senders or receivers) is fully gone/closed.

**Changes:**
- Add `BroadcastChannel` (tuple-style constructor returning an initial `BroadcastSender`/`BroadcastReceiver`) and rename broadcast endpoint classes to `BroadcastSender`/`BroadcastReceiver`.
- Add new sender capability interfaces (`SubscribableSender`, `ClonableSender`, `ClonableSubscribableSender`) and export them publicly.
- Deprecate `Broadcast` and `Anycast`, update release notes, and add tests covering broadcast auto-close behavior.

### DIFF
--- a/src/frequenz/channels/__init__.py
+++ b/src/frequenz/channels/__init__.py
@@ -80,7 +80,7 @@ Extra utility receivers:
 """
 
 from ._anycast import Anycast
-from ._broadcast import Broadcast
+from ._broadcast import Broadcast, BroadcastReceiver, BroadcastSender
 from ._exceptions import ChannelClosedError, ChannelError, Error
 from ._generic import (
     ChannelMessageT,
@@ -106,6 +106,8 @@ from ._sender import Sender, SenderClosedError, SenderError
 __all__ = [
     "Anycast",
     "Broadcast",
+    "BroadcastReceiver",
+    "BroadcastSender",
     "ChannelClosedError",
     "ChannelError",
     "ChannelMessageT",
--- a/src/frequenz/channels/__init__.py
+++ b/src/frequenz/channels/__init__.py
@@ -80,7 +80,7 @@ Extra utility receivers:
 """
 
 from ._anycast import Anycast
-from ._broadcast import Broadcast, BroadcastReceiver, BroadcastSender
+from ._broadcast import Broadcast, BroadcastChannel, BroadcastReceiver, BroadcastSender
 from ._exceptions import ChannelClosedError, ChannelError, Error
 from ._generic import (
     ChannelMessageT,
@@ -101,16 +101,26 @@ from ._select import (
     select,
     selected_from,
 )
-from ._sender import Sender, SenderClosedError, SenderError
+from ._sender import (
+    ClonableSender,
+    ClonableSubscribableSender,
+    Sender,
+    SenderClosedError,
+    SenderError,
+    SubscribableSender,
+)
 
 __all__ = [
     "Anycast",
     "Broadcast",
+    "BroadcastChannel",
     "BroadcastReceiver",
     "BroadcastSender",
     "ChannelClosedError",
     "ChannelError",
     "ChannelMessageT",
+    "ClonableSender",
+    "ClonableSubscribableSender",
     "Error",
     "ErroredChannelT_co",
     "LatestValueCache",
@@ -130,6 +140,7 @@ __all__ = [
     "SenderError",
     "SenderMessageT_co",
     "SenderMessageT_contra",
+    "SubscribableSender",
     "UnhandledSelectedError",
     "merge",
     "select",
--- a/src/frequenz/channels/__init__.py
+++ b/src/frequenz/channels/__init__.py
@@ -102,8 +102,8 @@ from ._select import (
     selected_from,
 )
 from ._sender import (
-    ClonableSender,
-    ClonableSubscribableSender,
+    CloneableSender,
+    CloneableSubscribableSender,
     Sender,
     SenderClosedError,
     SenderError,
@@ -119,8 +119,8 @@ __all__ = [
     "ChannelClosedError",
     "ChannelError",
     "ChannelMessageT",
-    "ClonableSender",
-    "ClonableSubscribableSender",
+    "CloneableSender",
+    "CloneableSubscribableSender",
     "Error",
     "ErroredChannelT_co",
     "LatestValueCache",
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -211,7 +211,7 @@ class Broadcast(Generic[ChannelMessageT]):
         """The number of senders attached to this channel."""
 
         self._receivers: dict[
-            int, weakref.ReferenceType[_Receiver[ChannelMessageT]]
+            int, weakref.ReferenceType[BroadcastReceiver[ChannelMessageT]]
         ] = {}
         """The receivers attached to the channel, indexed by their hash()."""
 
@@ -272,13 +272,13 @@ class Broadcast(Generic[ChannelMessageT]):
         """Close the channel, deprecated alias for `aclose()`."""  # noqa: D402
         return await self.aclose()
 
-    def new_sender(self) -> ClonableSubscribableSender[ChannelMessageT]:
+    def new_sender(self) -> BroadcastSender[ChannelMessageT]:
         """Return a new sender attached to this channel."""
-        return _Sender(self)
+        return BroadcastSender(self)
 
     def new_receiver(
         self, *, name: str | None = None, limit: int = 50, warn_on_overflow: bool = True
-    ) -> Receiver[ChannelMessageT]:
+    ) -> BroadcastReceiver[ChannelMessageT]:
         """Return a new receiver attached to this channel.
 
         Broadcast receivers have their own buffer, and when messages are not
@@ -294,7 +294,7 @@ class Broadcast(Generic[ChannelMessageT]):
         Returns:
             A new receiver attached to this channel.
         """
-        recv: _Receiver[ChannelMessageT] = _Receiver(
+        recv: BroadcastReceiver[ChannelMessageT] = BroadcastReceiver(
             self, name=name, limit=limit, warn_on_overflow=warn_on_overflow
         )
         self._receivers[hash(recv)] = weakref.ref(recv)
@@ -320,7 +320,7 @@ class Broadcast(Generic[ChannelMessageT]):
 _T = TypeVar("_T")
 
 
-class _Sender(ClonableSubscribableSender[_T]):
+class BroadcastSender(ClonableSubscribableSender[_T]):
     """A sender to send messages to the broadcast channel.
 
     Should not be created directly, but through the
@@ -400,9 +400,9 @@ class _Sender(ClonableSubscribableSender[_T]):
             self._channel._sender_count -= 1
 
     @override
-    def clone(self) -> _Sender[_T]:
+    def clone(self) -> BroadcastSender[_T]:
         """Return a clone of this sender."""
-        return _Sender(self._channel)
+        return BroadcastSender(self._channel)
 
     @override
     def subscribe(
@@ -410,7 +410,7 @@ class _Sender(ClonableSubscribableSender[_T]):
         name: str | None = None,
         limit: int = 50,
         warn_on_overflow: bool = True,
-    ) -> Receiver[_T]:
+    ) -> BroadcastReceiver[_T]:
         """Return a new receiver attached to this sender's channel.
 
         Args:
@@ -445,7 +445,7 @@ class _Sender(ClonableSubscribableSender[_T]):
         return f"{type(self).__name__}({self._channel!r})"
 
 
-class _Receiver(Receiver[_T]):
+class BroadcastReceiver(Receiver[_T]):
     """A receiver to receive messages from the broadcast channel.
 
     Should not be created directly, but through the
--- a/src/frequenz/channels/_broadcast.py
+++ b/src/frequenz/channels/_broadcast.py
@@ -391,6 +391,7 @@ class BroadcastSender(CloneableSubscribableSender[_T]):
         for _hash in stale_refs:
             del self._channel._receivers[_hash]
         if self._channel._auto_close_enabled and len(self._channel._receivers) == 0:
+            await self._channel.aclose()
             raise SenderError("The channel was closed", self) from ChannelClosedError(
                 self._channel
             )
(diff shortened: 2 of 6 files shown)


shsms:
Based on the one shot PR, which also needs more docs and examples.  

cc: @simonvoelcker

src/frequenz/channels/_broadcast.py
```diff
@@ -390,12 +449,12 @@ def __repr__(self) -> str:
...
-class _Receiver(Receiver[_T]):
+class BroadcastReceiver(Receiver[_T]):
     """A receiver to receive messages from the broadcast channel.
 
     Should not be created directly, but through the
-    [Broadcast.new_receiver()][frequenz.channels.Broadcast.new_receiver]
-    method.
+    [BroadcastSender.subscribe()][frequenz.channels.BroadcastSender.subscribe]
+    method of an existing sender.
     """
```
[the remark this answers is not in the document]
shsms: The docstrings are instructions for users of the library.  Internally, we have to do other things.

src/frequenz/channels/_broadcast.py
```diff
@@ -357,14 +381,18 @@ async def send(self, message: _T, /) -> None:
...
         stale_refs = []
         for _hash, recv_ref in self._channel._receivers.items():
             recv = recv_ref()
             if recv is None:
                 stale_refs.append(_hash)
                 continue
-            recv.enqueue(message)
+            recv._enqueue(message)
         for _hash in stale_refs:
             del self._channel._receivers[_hash]
```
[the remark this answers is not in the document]
shsms: Is true, fixed in 626bf13

src/frequenz/channels/_broadcast.py
```diff
@@ -541,3 +605,195 @@ def __repr__(self) -> str:
...
+    def __new__(
+        cls,
+        name: str,
+        resend_latest: bool = False,
+        limit: int = 50,
+        warn_on_overflow: bool = True,
+    ) -> BroadcastChannel[ChannelMessageT]:
+        """Create a new broadcast channel, deprecated alias for Broadcast."""
+        channel = Broadcast[ChannelMessageT](
+            name=name, resend_latest=resend_latest, auto_close=True
```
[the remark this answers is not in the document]
shsms: improved the docstring.

src/frequenz/channels/_broadcast.py
```diff
@@ -541,3 +605,195 @@ def __repr__(self) -> str:
...
+        )
```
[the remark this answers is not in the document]
shsms: I don't see any warnings.

src/frequenz/channels/_sender.py
```diff
@@ -109,3 +112,35 @@ def __init__(self, sender: Sender[SenderMessageT_co]):
...
+
+class SubscribableSender(Sender[SenderMessageT_contra], ABC):
+    """A [Sender][frequenz.channels.Sender] that can be subscribed to."""
+
+    @abstractmethod
+    def subscribe(self) -> Receiver[SenderMessageT_contra]:
+        """Subscribe to this sender.
+
+        Returns:
+            A new sender that sends messages to the same channel as this sender.
```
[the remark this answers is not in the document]
shsms: done

src/frequenz/channels/_broadcast.py
```diff
@@ -317,12 +334,12 @@ def __repr__(self) -> str:
...
-class _Sender(Sender[_T]):
+class BroadcastSender(ClonableSubscribableSender[_T]):
     """A sender to send messages to the broadcast channel.
 
     Should not be created directly, but through the
-    [Broadcast.new_sender()][frequenz.channels.Broadcast.new_sender]
-    method.
+    [BroadcastSender.clone()][frequenz.channels.BroadcastSender.clone]
+    method of an existing sender.
     """
```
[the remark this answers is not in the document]
shsms: The docs are for external users of the library, they need to be constucted somewhere, so they can be cloned by external users.

src/frequenz/channels/_broadcast.py
```diff
@@ -357,14 +381,18 @@ async def send(self, message: _T, /) -> None:
             raise SenderError("The channel was closed", self) from ChannelClosedError(
                 self._channel
             )
+        if self._channel._auto_close_enabled and len(self._channel._receivers) == 0:
```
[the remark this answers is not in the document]
shsms: No, receivers check for senders.  so it is fine.  `Broadcast` has the old aclose method only for backward compatibility.

src/frequenz/channels/_broadcast.py
```diff
@@ -492,6 +551,11 @@ async def ready(self) -> bool:
         while len(self._q) == 0:
             if self._channel._closed or self._closed:
                 return False
+            if self._channel._auto_close_enabled and (
+                self._channel._sender_count == 0 or len(self._channel._receivers) == 0
+            ):
```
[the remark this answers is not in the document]
shsms: done

src/frequenz/channels/_broadcast.py
```diff
@@ -379,7 +407,38 @@ async def aclose(self) -> None:
...
+            self._channel._sender_count == 0  # pylint: disable=protected-access
+            and self._channel._auto_close_enabled  # pylint: disable=protected-access
+        ):
+            await self._channel.aclose()
+
+    def __del__(self) -> None:
+        """Clean up this sender."""
+        if not self._closed:
+            self._channel._sender_count -= 1
+
```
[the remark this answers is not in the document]
shsms: Wrapped in try..except.

src/frequenz/channels/_sender.py
```diff
@@ -109,3 +112,35 @@ def __init__(self, sender: Sender[SenderMessageT_co]):
...
+        """
+
+
+class ClonableSubscribableSender(
+    SubscribableSender[SenderMessageT_contra],
+    ClonableSender[SenderMessageT_contra],
+    ABC,
+):
+    """A [Sender][frequenz.channels.Sender] that can be both cloned and subscribed to."""
```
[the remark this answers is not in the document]
shsms: fixed.

src/frequenz/channels/__init__.py
```diff
@@ -101,14 +101,26 @@
...
 )
-from ._sender import Sender, SenderClosedError, SenderError
+from ._sender import (
+    ClonableSender,
+    ClonableSubscribableSender,
+    Sender,
+    SenderClosedError,
+    SenderError,
+    SubscribableSender,
+)
```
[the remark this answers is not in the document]
shsms: fixed.

src/frequenz/channels/_broadcast.py
```diff
@@ -379,7 +407,53 @@ async def aclose(self) -> None:
...
+            self._channel._sender_count == 0  # pylint: disable=protected-access
+            and self._channel._auto_close_enabled  # pylint: disable=protected-access
+        ):
+            await self._channel.aclose()
+
+    def __del__(self) -> None:
+        """Clean up this sender."""
+        try:
+            if not self._closed:
+                self._channel._sender_count -= 1
```
llucax: This can probably fail because when this instance is deleted by the GC because there is no guarantee in order, `self._channel` might be already destroyed when this runs. I think we should probably leave the `__del__` out and required users to properly close receivers explicitly/using them as context managers. Also not calling the channel's `aclose()` could still leave things in a bad state.

src/frequenz/channels/_broadcast.py
llucax, line 608: Shouldn't this also trigger a closing of the channel if this is the last receiver? I guess to make this happen we need to transition to `aclose()` for receivers, right?

src/frequenz/channels/_broadcast.py
```diff
@@ -364,9 +388,13 @@ async def send(self, message: _T, /) -> None:
...
-            recv.enqueue(message)
+            recv._enqueue(message)
         for _hash in stale_refs:
             del self._channel._receivers[_hash]
+        if self._channel._auto_close_enabled and len(self._channel._receivers) == 0:
+            raise SenderError("The channel was closed", self) from ChannelClosedError(
+                self._channel
+            )
         async with self._channel._recv_cv:
             self._channel._recv_cv.notify_all()
```
llucax: Shouldn't `self._channel._latest = message` and `recv._enqueue(message)`be done after all the checks are done (right before we notify all receivers)? Otherwise it might be we report a send failed, but still receivers will receive the message.

llucax (commented):
I think there might be a few bugs left. If those are real bugs, it would be good to add test cases for them too.

src/frequenz/channels/_broadcast.py
shsms: I see, that's why copilot was suggesting getattr instead I guess.  It also recommended wrapping the whole thing under `try:except Exception:pass`, which I did.  So it might work or might not, but won't make things worse, right?

`aclose()` on channel wakes up all receivers, so they can stop waiting.  If all senders are gone, we call `channel.aclose()`: https://github.com/frequenz-floss/frequenz-channels-python/blob/ec05bfb6f657ebd22916b4420e8997812aa571a6/src/frequenz/channels/_broadcast.py#L415-L419

There's no other place where we need to call channel.aclose().  Not when all the receivers have gone.

src/frequenz/channels/_broadcast.py
shsms: No, the closing of the channel is really a closing of the receivers.  But if none are left, there's nothing for the channel close to do.

We might still consider `aclose` for receivers, but separate problem I guess.

src/frequenz/channels/_broadcast.py
shsms: If `self.channel.receivers` is already empty, there won't be a `enqueue` call.  There would just be a sender error.

If `self.channels.receivers` has only stale receivers, they would get cleaned up first, `enqueue` will not be called.  Then there would be a sender error.

If there are active receivers, `enqueue` will be called on them.  They are not stale, so they will stay.  No sender error will be raised.

The sendererror happens below, just to make sure we've dropped stale receivers first.

src/frequenz/channels/_broadcast.py
llucax: So, the "issue" here is, if we don't close the channel (don't mark it as closed), a new `subscribe()` can "resurrect" the channel. I'm not sure if this is intended behavior, it might be, but if it is we should document it more clearly. I'm not sure if it is a good idea though, because if we do this it means once you got `SenderClosedError` for a sender, a new `send()` might succeed, which sounds weird to me.

Suggested new test (that doesn't allow for resurrection, could be adapted if we want to allow it)
```py
async def test_broadcast_auto_close_cannot_be_resurrected_by_new_subscription() -> None:
    """Closing the last receiver should keep the channel closed."""
    sender, receiver = BroadcastChannel[int](name="auto-close-test")

    receiver.close()

    with pytest.raises(SenderError) as excinfo:
        await sender.send(1)
    assert isinstance(excinfo.value.__cause__, ChannelClosedError)

    late_receiver = sender.subscribe()  # <--- This should probably raise SenderError/ChannelClosedError

    with pytest.raises(SenderError) as excinfo:
        await sender.send(2)  # <--- This doesn't raise now
    assert isinstance(excinfo.value.__cause__, ChannelClosedError)
```

src/frequenz/channels/_broadcast.py
llucax: OK, so it is only about `_latest`. It seems updating `_latest` doesn't have a visible effect. This works as expected:

```py
async def test_broadcast_failed_send_does_not_update_latest_message() -> None:
    """A send that fails because the channel is closed should not mutate _latest."""
    sender, receiver = BroadcastChannel[int](
        name="auto-close-test",
        resend_latest=True,
    )
    await sender.send(1)
    await sender.aclose()
    with pytest.raises(SenderError):
        await sender.send(2)
    assert await receiver.receive() == 1
```

That said, it is weird to me that `_latest` is updated even if the send failed. Unless there is a very good reason to not move it to after it is really sent, I don't see why we shouldn't do it:

```py
async def test_broadcast_failed_send_does_not_update_latest_message() -> None:
    """A send that fails because the channel is closed should not mutate _latest."""
    sender, receiver = BroadcastChannel[int](
        name="auto-close-test",
        resend_latest=True,
    )
    await sender.send(1)
    assert await receiver.receive() == 1
    receiver.close()

    with pytest.raises(SenderError):
        await sender.send(2)

    assert sender._channel._latest == 1  # <--- fails because it was updated to 2 anyways
```

llucax: Unrelated to this, but I just noticed sometimes we raise `SenderClosedError` (`sender._closed is True`) and sometimes `SenderError` from `ChannelClosedError` (`sender._channel._closed is True`). I wonder if we should use `SenderClosedError` from `ChannelClosedError` instead, it sounds like if a channel is closed the sender should be closed too. I guess it relates to the question if senders/channels can be resurrected by re-subscribing or not.

That said, I guess this comes from before, and we are doing it like this in other places, so I'm OK to leave the broader question for another time.

src/frequenz/channels/_broadcast.py
llucax: This seems to be a real problem. The `__del__()` can still end up in blocked receivers:

This times out after a second:
```py
import gc

async def test_broadcast_receive_unblocks_when_last_sender_is_garbage_collected() -> None:
    """A blocked receiver should stop promptly when the last sender disappears."""
    sender, receiver = BroadcastChannel[int](name="auto-close-test")
    sender_2 = sender.clone()

    await sender.aclose()

    receive_task = asyncio.create_task(receiver.receive())
    await asyncio.sleep(0)

    # Adding await sender_2.aclose() here fixes it, of course
    del sender_2
    gc.collect()

    with pytest.raises(ReceiverStoppedError):
        async with asyncio.timeout(1):
            await receive_task
```

I would remove the `__del__`, it doesn't really help much and it is wrong. You should never access other mutable objects in the `__del__` anyway, it is intended only to free resources like from external C libraries or stuff like that, they are not proper destructors like in other languages).

shsms: ok, removed.

src/frequenz/channels/_broadcast.py
shsms: Fixed.  Also interesting that we allowed `new_receiver` calls from closed channels.  Won't be relevant soon though.

src/frequenz/channels/_broadcast.py
shsms: >  I don't see why we shouldn't do it

I wasn't objecting, I just wasn't aware of the issue.  I've moved the line updating `_latest` to after the raise.

shsms:
Ready again @llucax

simonvoelcker (approved):

llucax (approved):

PR #542, opened 2026-06-30
docs: fix and improve docstrings across all modules

llucax:
🤖 This pull request is part of an automated effort to improve the docstring documentation of all our projects.

Main issues being resolved:
- **Absolute → relative cross-references**: All internal cross-references now use dot notation (e.g. `[.method]`, `[..sibling]`, `[...package_symbol]`) enabled by the new `relative_crossrefs: true` mkdocstrings option. Absolute paths are kept only where relative refs cannot work consistently (class docstrings embedded in user-guide pages at a different context depth, and methods inherited across modules)
- **Bare backticks without links**: Replaced bare ```Symbol``` mentions with proper cross-references everywhere a symbol is linkable
- **Missing display-text backticks**: Added backticks to cross-reference display text (e.g. `[Symbol][path]` → `[`Symbol`][path]`)
- **Wrong summary mood**: Fixed boolean-returning methods from `"Check if/whether"` to `"Return whether"`; fixed property summaries from imperative to noun phrase
- **Grammar and typos**: Fixed `deliver` → `delivers`, `to sends` → `to send`, `a exception` → `an exception`, `seding` → `sending`, `massage` → `message`, `schduled` → `scheduled`, article fixes
- **Factual errors**: `send()` was described as returning `False` when channel is closed (it raises `SenderError`); wrong exception class `ReceiverClosedError` → `ReceiverStoppedError` in timer.py
- **Missing `Args:`/`Raises:` sections**: Added to `OneshotSender.send()` and `GroupingLatestValueCache` dict methods
- **Wrong example format**: Converted `# Example` heading to `Example:` admonition in `_oneshot.py`; fixed broken import path and assertion in `_pipe.py` example
- **mkdocs.yml config**: Added `relative_crossrefs: true` (convention rollout)
- **pyproject.toml config**: Added `check-class-attributes`, `check-style-mismatch`, `require-inline-class-var-docs` pydoclint options (convention rollout)

Other things to note:
- The `pytest_min` session was not run; all 121 tests pass in `pytest_max`.
- The `_pipe.py` example had a wrong import path (`Pipe` imported from `frequenz.channels` instead of `frequenz.channels.experimental`) and a wrong assertion (`== 11` → `== 10`); both are docstring code fixes.
- Relative refs cannot be used consistently in two contexts: (a) class docstrings in private modules (`_anycast.py`, `_broadcast.py`) that are embedded in user-guide pages get resolved at the private-module depth rather than the public API depth, so no single relative ref resolves correctly in both contexts; (b) methods inherited across modules (`Receiver.triggered`, `Receiver.map`, `Receiver.filter`) render at different depths depending on the inheriting class. Absolute paths are used in those specific cases.


shsms (commented):
You know you can ask AI to fixup things in place, right?  Then reviewing would have been easier.  Now there are 3 attempts/commits for each change.

But looks good otherwise.

shsms (approved):

